PR #569, opened 2022-10-18
New iso-remaster script and sample patcher scripts

ydirson:
Goal: replace most of the manual instructions in
https://xcp-ng.org/docs/develprocess.html#iso-modification

Base principle is "no need to be root", relies on fakeroot instead. Can use fuseiso+fuse-overlay if possible to avoid copies.

Supports applying changes to install.img, and to the remaining ISO contents.

Still needs better structuring to share more code between the two operating modes.

### DIFF
--- a/scripts/iso-remaster/iso-remaster.sh
+++ b/scripts/iso-remaster/iso-remaster.sh
@@ -1,0 +1,197 @@
+#! /bin/bash
+set -eE
+
+# TODO:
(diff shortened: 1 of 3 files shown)


stormi:
Looks good to me a first sight

stormi:
Note: I'm not for replacing manual instructions entirely yet. Rather to add instructions on how to use the script for those who want to use it, but also leave the manual instructions for whoever needs them and doesn't want to have to read a script to understand it.

PR #576, opened 2022-11-24
New creation scripts for installation iso

ydirson:


benjamreis (approved):

README-iso.md
```diff
@@ -0,0 +1,260 @@
...
+   * host contamination:
+     * ~~$B/etc/multipath.conf.disabled~~: not really, comes from BuildRequires
+  * metadata diffs
+    * removed locale files are not flagged as missing in CH8C install.img by "rpm -V", not clear now they achieved this
+    * CH8C has changes to filemodes (w and s bits removed), visible with rpm -V.  No clue what they tried to achieve.
+
+
+## from Samuel
+
+Comparison of install.img between XCP-ng 8.2.0's official ISOs and newly generated install.img
```
benjamreis, line 223: Do we want to keep this title?

README-iso.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+UEFI:
+
+```
+qemu-system-x86_64 -serial stdio -m 1G -cdrom xcp-ng-8.3-install.iso \
+  --bios /usr/share/edk2/ovmf/OVMF_CODE.fd -net none
+```
+
+### testing that scripts run correctly
```
benjamreis, line 115: Can this be launched in a GH action?

scripts/create-install-iso.sh
```diff
@@ -0,0 +1,262 @@
...
+   "$SCRATCHDIR/syslinux/usr/share/syslinux/isolinux.bin" \
+   "$SCRATCHDIR/syslinux/usr/share/syslinux/mboot.c32" \
+   "$SCRATCHDIR/syslinux/usr/share/syslinux/menu.c32" \
+   \
+   $ISODIR/boot/isolinux/
+
+
+## create final ISO
+
+if false; then
```
benjamreis, line 211: Why?

README-iso.md
ydirson: They do take some time, so we likely don't want to run the whole of them in GH.  To be discussed.

stormi (commented):
First comments after a commit by commit review, without actually reviewing the code of the scripts.

README-iso.md
```diff
@@ -0,0 +1,260 @@
...
+- development RPMs
+- debugging-symbols RPMs
+
+### configuration layers and package repositories
+
+Configuration layers are defined as a subdirectory of the `configs/`
+directory.  Commands are given a layer search path as
+`<base-config>[:<config-overlay>]* `.
+
+Standard layers are organized such two of the standard layers must be
```
stormi, line 44:
```suggestion
Standard layers are organized such as two of the standard layers must be
```

README-iso.md
```diff
@@ -0,0 +1,260 @@
...
+
+prove tests/ :: -l
+```
+
+
+## still to be done (a bit outdated but not that much)
+
+* [ ] .treeinfo files have an unsubst'd `{{timestamp}}`
+* [x] how to deal with the fact we can only reproduce 8.x.0 and 8.x.updates
+* [ ] (^^ linked ^^) customizability of available repos
```
stormi, line 160: Done, or there's more to be done about it?

iso/8.2/.treeinfo
```diff
@@ -1,17 +1,17 @@
...
+version = 3.2.1
 
 [branding]
 name = XCP-ng
-version = 8.2.0
+version = 8.2
 
 [build]
-number = release/stockholm/master/7
+#number = release/yangtze/master/58
```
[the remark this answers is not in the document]
stormi: Remains the question of how to differenciate a test build from an official release, and differenciate two subsequent builds. The timestamp will help to some extent, but if we want to go further we could maybe append stuff to either the build number (if diverging from the actual product build number which is in the `xcp-ng-release` RPM is not a problem) or the general name, based on the repo config used and the date.

For example, if we use a `8.3:ci` config, then `general.name` could become `XCP-ng-8.3-ci-20230517`.

Only builds that correspond to official releases would not bear this suffix, so `general.name` would remain `XCP-ng-8.3` for the official release. 

Ideally we'd use the same suffix for the produced ISOs, and this would be a `--name-suffix` setting who could have one the following values:
* `auto`: let the script build the suffix (default?)
* `none`: no suffix at all (official releases)
* anything else: use this input as the suffix. If I want to suffix it `-stormi-test-build-for-I225-V`, I can.

scripts/mirror-repos.sh
```diff
@@ -0,0 +1,19 @@
...
+DIST="$1"
+maybe_set_srcurl "$DIST"
+
+command -v lftp >/dev/null || die "required tool not found: lftp"
+
+lftp -c mirror \
+     --verbose \
+     --delete \
+     --exclude="/Source/|-debuginfo-|-devel[-_]" \
+     $SRCURL ~/mirrors/xcpng/$DIST
```
stormi, line 19: This path should be configurable. For example, I wouldn't put mirrors in this directory on my own system.

installimg/8.2.updates/etc/systemd/system/sshd.service.d/installer.conf
```diff
@@ -0,0 +1,5 @@
+[Unit]
```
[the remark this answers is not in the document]
ydirson: My choice to leave it here was because it is literally zero-cost to get the file as part of the file copying, whereas going to the SRPM requires extra work that will eventually be reverted, for very little benefit.

installimg/8.2.updates/etc/systemd/system/getty@tty2.service
```diff
@@ -0,0 +1,41 @@
+#  This file is based on the service file from systemd.
```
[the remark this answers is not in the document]
ydirson: This service has literally nothing to do with the installer, it is really rather part of the "installimg distro", just to have a getty running.  E.g. if we were to replace `host-installer` by another program, it would still fill its role.

README-iso.md
ydirson: `grub-mkrescue` is "the reference" in someway, providing largest platform support for booting, but OTOH adds tons of stuff we don't need at all. It was used as a reference to implement UEFI boot, and I kept it to have it handy if/when we need it, since the command options are not that obvious.  Eg. we may want to add support for x86 Macs some day.
Will add a comment.

scripts/create-install-iso.sh
ydirson: grub-mkrescue is "the reference" in someway, providing largest platform support for booting, but OTOH adds tons of stuff we don't need at all. It was used as a reference to implement UEFI boot, and I kept it to have it handy if/when we need it, since the command options are not that obvious. Eg. we may want to add support for x86 Macs some day.
Will add a comment.

README-iso.md
```diff
@@ -60,6 +60,10 @@ https://updates.xcp-ng.org/ and most of them are available through
 standard "repo" layers; e.g. the `testing` repository for `8.2` LTS can
 be used as `8.2:testing`.
 
+Custom repositories can be added with `--define-repo` flag.  They will
+be used by `yum` using the first `CUSTOMREPO.tmpl` template found in
+the layer search path (one is provided in `base`).
```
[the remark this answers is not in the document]
ydirson: Yes - right, the plural on "repositories" may not be enough to make that crystal clear :D

configs/staging/yum-repos.conf.tmpl
```diff
@@ -0,0 +1,6 @@
+[xcpng-staging]
```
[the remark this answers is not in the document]
ydirson: that's why it was the last commit ;)

README-iso.md
ydirson: yes

README-iso.md
ydirson: no :)

.gitignore
```diff
@@ -1,8 +1,12 @@
...
 /tests/test-results/
 /tests/trash directory.*
 
 # temporary files during image generation
 /tmpdir-*/
 /rootfs-*
 /yum-*.conf
+/installiso.*
+/tmp.*
+/bootx64-*.efi
```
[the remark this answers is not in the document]
ydirson: Especially as we now override `TMPDIR`, that looks like a great place to put everything.  Also means it won't be trivial any more to disable cleaning of one file type during debugging, but having everything in a single place overweights this.

README-iso.md
ydirson: right

scripts/mirror-repos.sh
ydirson: right

iso/8.2/.treeinfo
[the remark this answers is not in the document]
ydirson: If we start adding `%d%m%y` in filename, output file is not clobbered by default any more, so we'll likely want to have `%H%M%S` too to have a consistent behaviour and less ambiguity.

> * If I want to suffix it `-stormi-test-build-for-I225-V`, I can.

You can already by selecting a full name. Leaving the filename as a caller choice decouples implementation from policy, which is generally a good idea.  Probably `create-installimg` should not use a default output filename anyway like `create-install-iso`, and both should refuse toclobber unless some `--force-overwrite` is given.

.gitignore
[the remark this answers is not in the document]
ydirson: Done

configs/base/CUSTOMREPO.tmpl
```diff
@@ -0,0 +1,6 @@
+[@@REPOID@@]
+name=xcpng - Custom @@REPOID@@
+baseurl=@@REPOURL@@/@@RPMARCH@@/
+#gpgkey=https://xcp-ng.org/RPM-GPG-KEY-xcpng
```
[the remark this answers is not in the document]
stormi: Let's plan this for after the first merge.

configs/candidates/INCLUDE
```diff
@@ -0,0 +1 @@
+base
```
stormi, line 1: I think you inverted updates and candidates.

Expected:
* `testing` requires `candidates`
* `candidates` requires `updates`
* `updates` requires `base`

ydirson: fixed

configs/base/CUSTOMREPO.tmpl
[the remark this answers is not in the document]
ydirson: Since that will change the UI, will already change the syntax to use `!` separator now, and it can be extended later

iso/8.2/.treeinfo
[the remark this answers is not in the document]
ydirson: change done, no default output filename any more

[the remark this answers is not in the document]
ydirson: `.treeinfo` content generation from branding has been added to TODO list already

configs/base/CUSTOMREPO.tmpl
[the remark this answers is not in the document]
ydirson: added the key specification to TODO list

stormi (approved):
Looks good. Let's squash the fixups, and then move all of this to a new repository.

PR #581, opened 2022-12-08
Fixes and improvents to iso-remaster script and sample hooks

ydirson:


stormi (approved):
Minor comment: is the word "path" missing from the second commit title?

PR #598, opened 2023-04-04
iso-remaster: use createrepo_c in sample script

ydirson:
Plain `createrepo` is not available in Debian, using `createrepo_c` makes things more portable.


stormi (approved):
And AFAIK createrepo_c is generally preferred over createrepo: https://bugzilla.redhat.com/show_bug.cgi?id=1002395

PR #607, opened 2023-05-11
Add microsemi-smartpqi-alt to package provenance

gduperrey:


stormi (changes requested):
It is also needed in data/8.3

stormi (approved):

PR #608, opened 2023-05-16
Two scripts to automate iso generation in Jenkins

gduperrey:
gen-iso.sh is meant to generate the ISO files with the dev env with the docker container.

The manage-iso scp the isos into a sr iso and into the server. It manages the retention too, and uncompress the
ISOs into installer directories on the pxe.

### DIFF
--- a/scripts/gen-iso.sh
+++ b/scripts/gen-iso.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+set -ueE
+
+usage()
+{
+    echo "----------------------------------------------------------------------------------------"
+    echo "- The purpose of this script is to be executed inside the container of our development environment to generate ISO files for XCP-ng."
+    echo "- Usage: $0 <XCP-ng version> <REPOSITORY>"
+    echo "- All options are mandatory."
+    echo "- VERSION of XCP-ng: 8.2 or 8.3"
+    echo "- REPOSITORY: update, testing, ci, ..."
+    echo "- Example: $0 8.3 testing"
+    echo "----------------------------------------------------------------------------------------"
+}
+
+if [ $# -lt 2 ]; then
+    echo "Missing argument(s)"
+    echo
+    usage
+    exit 1
+fi
+
+VERSION="$1"
+
+if [ "$VERSION" != "8.3" ] && [ "$VERSION" != "8.2" ]; then
+    echo "Unsupported version. Please choose between 8.2 and 8.3."
+    exit
+fi
+
+REPOSITORY="$2"
+
+THEDATE=$(date +%Y%m%d)
+
+set -x
+
+NAMEIMG="install-${VERSION}-${REPOSITORY}.img"
+NAMEISO="xcp-ng-${VERSION}-${REPOSITORY}-nightly-${THEDATE}.iso"
+NAMEISONI="xcp-ng-${VERSION}-${REPOSITORY}-netinstall-nightly-${THEDATE}.iso"
+
+if [ "$VERSION" = "8.3" ]; then
+    MNTVOL="XCP-NG_83"
+elif [ "$VERSION" = "8.2" ]; then
+    MNTVOL="XCP-NG_82"
+else
+    MNTVOL="XCP-NG"
+fi
+
+sudo yum install -y genisoimage syslinux grub-tools createrepo_c
+
+cd /data
+
+sudo ./scripts/create-installimg.sh --srcurl "https://updates.xcp-ng.org/8/${VERSION}" -o "${NAMEIMG}" "${VERSION}":"${REPOSITORY}"
+
+./scripts/create-install-iso.sh --netinstall --srcurl "https://updates.xcp-ng.org/8/${VERSION}" -V "${MNTVOL}" -o "${NAMEISONI}" "${VERSION}":"${REPOSITORY}" "${NAMEIMG}"
+
+./scripts/create-install-iso.sh --srcurl "https://updates.xcp-ng.org/8/${VERSION}" -V "${MNTVOL}" -o "${NAMEISO}" "${VERSION}":"${REPOSITORY}" "${NAMEIMG}"
--- a/scripts/manage-isos.sh
+++ b/scripts/manage-isos.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+set -ueE
+
+usage()
+{
+    echo "----------------------------------------------------------------------------------------"
+    echo "- This script is for internal use at Vates to transfer and manage ISOs generated automatically by our Jenkins server and to store them on our PXE server and ISO SR."
+    echo "- Usage: $0 <XCP-ng version> <Repository> <PXE Server> <Path on the PXE server> <ISO SR Server> <Path on the ISO SR Server> <Path on the PXE Server for the installer>"
+    echo "- All options are mandatory."
+    echo "- Repository: updates, ci, testing..."
+    echo "- Version of XCP-ng: 8.2 or 8.3"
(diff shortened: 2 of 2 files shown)


scripts/manage-isos.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+
+set -u
+
+usage()
+{
+    echo "----------------------------------------------------------------------------------------"
+    echo "- Usage: $0 [version XCP-ng] [Repository] [Server PXE] [Path on the server PXE] [Server SR ISO] [Path on the server of the SR ISO] [Path on the pXE for th installer]"
```
ydirson, line 8: "PXE server", "ISO SR"

scripts/manage-isos.sh
```diff
@@ -0,0 +1,90 @@
...
+    echo "- All options are mandatory."
+    echo "- Repository: updates, ci, testing..."
+    echo "- Version of XCP-ng: 8.2 or 8.3"
+    echo "- Example: bash manage-isos.sh 8.3 myuser@my_pxe /my/path/pxe/ iso_user@srv-sr-iso /my/path/on/sriso/ /my/path/pxe/installers/"
+    echo "----------------------------------------------------------------------------------------"
+}
```
ydirson, line 14: A short description of what the command does would be great, especially since "manage" is a very generic word

scripts/manage-isos.sh
```diff
@@ -0,0 +1,90 @@
...
+    if [ -d ${NAME_LOCAL_DIRECTORY} ]; then
+        rm -Rf ${NAME_LOCAL_DIRECTORY}
+    fi
+fi
+
+#Manage the monthly one on the first of each month
+if [[ $(date +%d) == 01 ]]; then
+    echo "We're the first of the month, so we create a monthly iso to keep 3 months."
+    NEW_NAME_ISO=${FILE_NAME_ISO/"nightly"/"monthly"}
+    scp ${FILE_NAME_ISO} ${SERVERPXE}:${PATHPXE}${NEW_NAME_ISO}
```
ydirson, line 75: Since everything is on a single drive, a hardlink would save a bit of disk space and transfer time (bonus: `ls -l` in nightly directory would show that a nightly is also a monthly)

scripts/manage-isos.sh
nagilum99: "for th installer" -> "for the installer"

scripts/gen-iso.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "- VERSION of XCP-ng: 8.2 or 8.3"
+    echo "- REPOSITORY: update, testing, ci, ..."
+    echo "- Example: $0 8.3 testing"
+    echo "----------------------------------------------------------------------------------------"
+}
+
+
+VERSION="$1"
+
+if [ $VERSION != "8.3" ] && [ $VERSION != "8.2" ]; then
```
ydirson, line 20: variables like `VERSION` which are built from user-provided input should also be protected from special-chars-in-args, using double quotes (paranoid approach is "always quote variable expansion to be sure")

```
[builder@1c3fd59db0dd xcp]$ ./scripts/gen-iso.sh "8.1 != 8.1 -a 8.3" update
+ VERSION='8.1 != 8.1 -a 8.3'
+ '[' 8.1 '!=' 8.1 -a 8.3 '!=' 8.3 ']'
+ REPOSITORY=update
++ date +%Y%m%d
...
+ sudo ./scripts/create-installimg.sh --srcurl http://mirrors.xcp-ng.org/8/8.1 '!=' 8.1 -a 8.3 -o install-8.1 '!=' 8.1 -a 8.3-update.img 8.1 '!=' 8.1 -a 8.3:update
```

scripts/manage-isos.sh
```diff
@@ -0,0 +1,93 @@
...
+
+#get files names
+ISO_FILE_NAME_NI=$(ls *.iso | grep netinstall)
+ISO_FILE_NAME=$(ls *.iso | grep -v netinstall)
+
+#scp to the pxe server
+echo "We're doing the scp to the pxe server for the two isos."
+scp *.iso ${PXESERVER}:${PXEPATH}
+echo "Creating or changing the link latest on the pxe."
+ssh ${PXESERVER} 'ln -sf '${PXEPATH}${ISO_FILE_NAME}' '${PXEPATH}'xcp-ng-'${VERSION}'-'${REPO}'-latest'
```
ydirson, line 53: A single set of double-quotes would make the line more readable (like was done in other places), and would allow to single-quote each variable argument to `ln` to also protect that one against special-chars-in-args.

scripts/manage-isos.sh
```diff
@@ -0,0 +1,93 @@
...
+    ssh ${PXESERVER} 'ln '${PXEPATH}${ISO_FILE_NAME} ${PXEPATH}${MONTHLY_ISO_NAME}
+    ssh ${PXESERVER} "find ${PXEPATH} -type f -mtime +182 -name '*monthly*.iso' -delete"
+fi
+
+#scp to the SR ISO
+echo "Copy of the isos files to the SR ISO."
+scp *.iso ${ISOSR}:${ISOSRPATH}
+ssh ${ISOSR} "find ${ISOSRPATH} -mtime +2 -name '*nightly*.iso' -delete"
+
+#clean files
```
ydirson, line 87: those would not be cleaned up if the script exited early because of an error, or user hitting ctrl-c
one way to deal with that is to setup early an exit hook like
```
trap 'rm -rf "$LOCAL_DIRECTORY_NAME" *.img *.iso' EXIT INT
```

scripts/manage-isos.sh
```diff
@@ -0,0 +1,93 @@
...
+echo "We clean the temp directory, if it exists."
+if [ -d ${LOCAL_DIRECTORY_NAME} ]; then
```
ydirson, line 89: `test -d` is unnecessary with `rm -f`

ydirson (commented):
A few remarks about generic shell-script robustness stuff, otherwise lgtm

scripts/manage-isos.sh
gduperrey: I prefer to keep it, so i only do the command if the directory exist.

scripts/gen-iso.sh
gduperrey: I don't understand that part.

But I'd like to recall this script isn't meant to be launch directly by an external user. It's mean to be launch by our Jenkins server for internal use. I think we can not insert errors in the arguments.

And the test check the variable. Si if it's something else, it will exit.

stormi: It just means you should always add `""` around a variable that you test, like in line 20.
```suggestion
if [ "$VERSION" != "8.3" ] && [ "$VERSION" != "8.2" ]; then
```

Be there an actual risk or not in the way the script is used, it's good practice and a good habit to take.

scripts/manage-isos.sh
```diff
@@ -0,0 +1,99 @@
...
+fi
+
+#get files names
+ISO_FILE_NAME_NI=$(ls *.iso | grep netinstall)
+ISO_FILE_NAME=$(ls *.iso | grep -v netinstall)
+
+function cleanup_files () {
+    #clean files
+    echo "We clean the temp directory, if it exists."
+    if [ -d ${LOCAL_DIRECTORY_NAME} ]; then
```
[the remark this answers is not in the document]
gduperrey: There's a lot of commands before, so if it failed the trap won't be executed as it is too far in the script. (If i understood right ho wthe trap works).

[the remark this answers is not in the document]
gduperrey: The function doesn't depend only on this variable. It's here to clean .img and .iso files too.

gduperrey:
Tests on our jenkins server are ok.

stormi (approved):

PR #614, opened 2023-07-06
Add nbd to package_provenance.csv for 8.2

gduperrey:


ydirson:
The point of this commit is not clear, this is the "provenance" list but from the commit message I expected to get the package added to the install

stormi:
Indeed, should be something like "Add nbd to package_provenance.csv for 8.2"

gduperrey:
Commit message changed with the one provided by Samuel.

PR #622, opened 2023-08-31
Change the name of the Volume ID in the iso generated

gduperrey:
To have a more explicit name for the Volume ID, we've changed the parameter to have a name like:
"XCP-ng version repository date"
Example: XCP-ng 8.3 updates 20230621

### DIFF
--- a/scripts/gen-iso.sh
+++ b/scripts/gen-iso.sh
@@ -38,13 +38,7 @@ NAMEIMG="install-${VERSION}-${REPOSITORY}.img"
 NAMEISO="xcp-ng-${VERSION}-${REPOSITORY}-nightly-${THEDATE}.iso"
 NAMEISONI="xcp-ng-${VERSION}-${REPOSITORY}-netinstall-nightly-${THEDATE}.iso"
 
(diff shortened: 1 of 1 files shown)


benjamreis (approved):

benjamreis:
little typo in the title of the PR and the commit :)

ydirson (changes requested):
spaces and lower-case are not supported, and the string length is limited

gduperrey:
@ydirson Yes we nknow that the string lenght is limited. It's 32. That's why we change it from what was in the kanban card.

And there isn't any problem with space and lower case on our tests: `Volume id: XCP-ng 8.2 ci 20230829`

ydirson:
> And there isn't any problem with space and lower case on our tests: `Volume id: XCP-ng 8.2 ci 20230829`

Did you check how Windows deals with it?

gduperrey:
What do you mean ?
I haven't any windows right now.

ydirson:
A Windows VM should do the job :)

ydirson (approved):

PR #651, opened 2024-06-18
Fix alternate upgrade method of XCP-ng

benjamreis:
The plugin `prepare_host_upgrade.py` looks for the added entries in the `.treeinfo` in order to do the upgrade.

### DIFF
--- a/iso/8.3/.treeinfo
+++ b/iso/8.3/.treeinfo
@@ -1,3 +1,11 @@
+[platform]
+name = XCP
+version = 3.4.0
+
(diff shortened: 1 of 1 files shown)


stormi (commented):
With the new ISO generation scripts, the file that is actually used is https://github.com/xcp-ng/create-install-image/blob/master/templates/iso/8.3/.treeinfo

benjamreis:
Replaced by: https://github.com/xcp-ng/create-install-image/pull/23

PR #652, opened 2024-06-18
iso-remaster: improve messages

ydirson:


stormi (approved):
LGTM. First commit has a title a few characters too long for tools with limited width, so it would be better if it were shorter.

PR #654, opened 2024-07-08
iso-remaster: a bit of doc

ydirson:


gthvn1 (approved):
tested on my fedora40 and it worked (only need to remove `cdrkit-isotools` that is not available). Otherwise it looks ok.

stormi (approved):

PR #665, opened 2024-10-14
New script to setup an local repo for offline updates

stormi:
With this script, one can setup a local repository from an archive which contains the files of a repository, and use it to install updates or additional packages.

The same script is meant to be used to update the repository. In order not to use too much space on the disk, we accept that the operation makes the repository unusable for the duration of the local copy.

The archive must contain a top level directory named after the repo type. For the `updates` repository, the script will look for `^updates/$`.

Though I haven't produced any archive to test it, this is supposed to be usable not only for release updates, but also for other repos for testing, such as `testing` or `candidates`, if the need arises later (I bet it will).


scripts/setup_offline_updates.sh
```diff
@@ -0,0 +1,61 @@
...
+if ! tar -tf "$TAR_FILE" | grep -q "^${REPO_TYPE}/$"; then
+    echo "Error: directory $REPO_TYPE/ not found in archive $TAR_FILE."
+    exit 3
+fi
+
+mkdir -p "$REPOS_DIR"
+
+if [ -d "$REPOS_DIR/$REPO_TYPE" ]; then
+    echo "- Deleting old repo at $REPOS_DIR/$REPO_TYPE"
+    rm "$REPOS_DIR/$REPO_TYPE" -r
```
ydirson, line 41: Maybe add `-f` to be sure, possibly subject to a `-f` flag on this script?

scripts/setup_offline_updates.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+echo "- Extracting $TAR_FILE to $REPOS_DIR/$REPO_TYPE"
+tar -xf "$TAR_FILE" -C "$REPOS_DIR"
+
+echo "- Writing /etc/yum.repos.d/$REPONAME.repo"
```
ydirson, line 47: Those will live alongside `xcp-ng.repo`.  Maybe they should be named `xcp-ng-$REPONAME.repo`, and those files could also be used in the product instead of `xcp-ng.repo`?

scripts/setup_offline_updates.sh
```diff
@@ -0,0 +1,61 @@
...
+cat <<EOF > "/etc/yum.repos.d/$REPONAME.repo"
+[xcp-ng-offline-$REPO_TYPE]
+name=XCP-ng Offline ${REPO_TYPE^} Repository
+baseurl=file://$REPOS_DIR/$REPO_TYPE/x86_64/
+enabled=1
+gpgcheck=1
```
ydirson, line 53: any reason not to use `repo-gpgcheck`?

scripts/setup_offline_updates.sh
```diff
@@ -0,0 +1,61 @@
...
+TAR_FILE="$2"
+REPOS_DIR="/var/local/xcp-ng-repos"
+REPONAME="xcp-ng-offline-$REPO_TYPE"
+
+if [ ! -e "$TAR_FILE" ]; then
+    echo "Error: file $TAR_FILE not found."
+    exit 2
+fi
+
+if ! tar -tf "$TAR_FILE" | grep -q "^${REPO_TYPE}/$"; then
```
ydirson, line 32: truncated/corrupted tar file detection needs `set -o pipefail`

stormi: Well, the story with `pipefail` is funny. `grep -q` stops early as soon as it finds a match, which sends SIGPIPE to tar, thus we end with return code 141 and the whole condition with `!` never evaluates to false.

If the tar file is corrupted, we'll have an error when we extract it anyway, right?

scripts/setup_offline_updates.sh
stormi: I'm not sure what you mean. The file name is already `xcp-ng-offline-updates.repo` in the case of `$REPO_TYPE` being `updates`.

scripts/setup_offline_updates.sh
stormi: I'd rather have the script fail if `rm -r` fails for some reason, to have an occasion to understand why it happened.

scripts/setup_offline_updates.sh
ydirson: > If the tar file is corrupted, we'll have an error when we extract it anyway, right?

Yes, but we'll have removed the old repo already

scripts/setup_offline_updates.sh
ydirson: That's why I was suggesting to let the user specify `-f` if needed

scripts/setup_offline_updates.sh
ydirson: Oh, I missed that.
The 2nd part of the remark was about whether we really want to keep offline repos *in addition to* networked ones (which require the offline users to disable the networked ones).  Wouldn't it be better to just replace the refs to remote repos with a ref to the offline ones?

scripts/setup_offline_updates.sh
ydirson: Clearly looks like I'm missing context, then

scripts/setup_offline_updates.sh
stormi: I haven't put it in the script, but I added the instructions to disable the existing ones at initial pool setup in the related doc I'm writing.

scripts/setup_offline_updates.sh
stormi: It's an update repo which only contains the latest of each package, and I haven't signed the new repodata, but I could if we think it's useful.

scripts/setup_offline_updates.sh
stormi: I'll test the tarball integrity first then.

scripts/setup_offline_updates.sh
stormi: Maybe for AirGapped v2 (the next iteration). If I need to add non positional arguments, I'll switch to python with argparse :)

scripts/setup_offline_updates.sh
stormi: Actually I do produce signatures for the repo metadata :sweat_smile:. I added `repo_gpgcheck`.

stormi:
I have entirely rewritten the script and changed its behaviour. It's been tested internally. Due to timing constraints, I have to merge the PR now.

PR #676, opened 2025-02-10
Port rpmwatcher to python3

ydirson:


stormi:
It fails when it needs to run a script from within a CentOS 7.5 container:
```
+ docker run --rm -t --privileged --network host -v /home/rpmwatcher/data:/data -v /home/rpmwatcher/git/xcp/scripts/rpmwatcher:/scripts centos:7.5.1804 /scripts/rpmwatcher_extract_deps.py 8.3 /data
/bin/env: python3: No such file or directory
```

ydirson:
Maybe it is time to add `python3` to the default build env, with a goal of getting rid of `python2` ASAP (maybe even right now)?

stormi:
I don't know if it changes much to the build env. The container here is a basic centos 7.5 container (to which local XCP-ng repos are then added). It resembles xcp-ng-build-env but it's not it.

rzr:
This change has been stalling, is it ready or should it be set to draft ?

stormi:
It can become a draft but we'll have to finish it later.

PR #677, opened 2025-02-11
Generation of a RPM status dashboard to compare with xs8

ydirson:


ydirson:
Added type annotations

ydirson:
Added `candidates` and prevent 1st column wrapping

PR #679, opened 2025-02-28
add `--test-build` flag to `koji_build`

glehmann:
When run with `koji_build.py --test-build gln --nowait v8.2-u-gln1 .`, it creates a new commit to bump the release and push it to github.
That way the release `2.5%{?dist}` becomes `2.5.0.gln.1%{dist}`.

The branch pushed to github is named `koji/test/<test build name>/<date>` (for example `koji/test/gln/2025-02-28-11-48-40`) and cleanup when it is more than 3 hours old the next time someone uses koji_build on that repository.

Note that I've use an extra package — `specfile` that must be installed with `pip install --user specfile` before using the script.

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -112,6 +112,8 @@ def main():
         urls = []
         for d in git_repos:
             remote, hash = get_repo_and_commit_info(d)
+            if test_build:
+                hash = push_bumped_release(d, test_build)
             urls.append(koji_url(remote, hash))
         command = ['koji', 'chain-build', target] + (' : '.join(urls)).split(' ') +  (['--nowait'] if is_nowait else [])
         print('  '.join(command), flush=True)
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -73,7 +73,7 @@ def main():
         remote, hash = get_repo_and_commit_info(git_repos[0])
         url = koji_url(remote, hash)
         command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
-        subprocess.check_call(['echo'] + command)
+        print('  '.join(command), flush=True)
         subprocess.check_call(command)
     else:
         urls = []
@@ -81,7 +81,7 @@ def main():
             remote, hash = get_repo_and_commit_info(d)
             urls.append(koji_url(remote, hash))
         command = ['koji', 'chain-build', target] + (' : '.join(urls)).split(' ') +  (['--nowait'] if is_nowait else [])
-        subprocess.check_call(['echo'] + command)
+        print('  '.join(command), flush=True)
         subprocess.check_call(command)
 
 if __name__ == "__main__":
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -77,7 +77,22 @@ def clean_old_branches(git_repo):
             print("removing outdated remote branch(es)", flush=True)
             subprocess.check_call(['git', 'push', '--delete', 'origin'] + old_branches)
 
-def push_bumped_release(git_repo, test_build_id):
+def xcpng_version(target):
+    xcpng_version_match = re.match(r'^v(\d+\.\d+)-u-\S+$', target)
+    if xcpng_version_match is None:
+        raise Exception(f"Can't find XCP-ng version in {target}")
+    return xcpng_version_match.group(1)
+    
+def find_next_build_number(package, spec, target, test_build_id):
+    builds = subprocess.check_output(['koji', 'list-builds', '--quiet', '--package', package]).decode().splitlines()
+    base_nvr = f'{package}-{spec.version}-{spec.release}.0.{test_build_id}.'
+    # use a regex to match %{macro} without actually expanding the macros
+    base_nvr_re = re.escape(re.sub('%{.+}', "@@@", base_nvr)).replace('@@@', '.*') + r'(\d+)' + re.escape(f'.xcpng{xcpng_version(target)}')
+    build_matches = [re.match(base_nvr_re, b) for b in builds]
+    build_nbs = [int(m.group(1)) for m in build_matches if m]
+    return sorted(build_nbs)[-1] + 1 if build_nbs else 1
+    
+def push_bumped_release(git_repo, target, test_build_id):
     t = datetime.now().strftime(TIME_FORMAT)
     branch = f'koji/test/{test_build_id}/{t}'
     with cd(git_repo), local_branch(branch):
@@ -87,14 +102,8 @@ def push_bumped_release(git_repo, test_build_id):
         with Specfile(spec_path) as spec:
             # find the next build number
             package = Path(spec_path).stem
-            builds = subprocess.check_output(['koji', 'list-builds', '--package', package]).decode().splitlines()[2:]
-            base_nvr = f'{package}-{spec.version}-{spec.release}.0.{test_build_id}.'
-            # use a regex to match %{macro} without actually expanding the macros
-            base_nvr_re = re.escape(re.sub('%{.+}', "@@@", base_nvr)).replace('@@@', '.*') + r'(\d+)'
-            build_matches = [re.match(base_nvr_re, b) for b in builds]
-            build_ids = [int(m.group(1)) for m in build_matches if m]
-            next_build_id = sorted(build_ids)[-1] + 1 if build_ids else 1
-            spec.release = f'{spec.release}.0.{test_build_id}.{next_build_id}'
+            next_build_nb = find_next_build_number(package, spec, target, test_build_id)
+            spec.release = f'{spec.release}.0.{test_build_id}.{next_build_nb}'
         subprocess.check_call(['git', 'commit', '--quiet', '-m', "bump release for test build", spec_path])
         subprocess.check_call(['git', 'push', 'origin', f'HEAD:refs/heads/{branch}'])
         commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode().strip()
@@ -131,7 +140,7 @@ def main():
         clean_old_branches(git_repos[0])
         remote, hash = get_repo_and_commit_info(git_repos[0])
         if test_build:
-            hash = push_bumped_release(git_repos[0], test_build)
+            hash = push_bumped_release(git_repos[0], target, test_build)
         url = koji_url(remote, hash)
         command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
         print('  '.join(command), flush=True)
@@ -142,7 +151,7 @@ def main():
             clean_old_branches(d)
             remote, hash = get_repo_and_commit_info(d)
             if test_build:
-                hash = push_bumped_release(d, test_build)
+                hash = push_bumped_release(d, target, test_build)
             urls.append(koji_url(remote, hash))
         command = ['koji', 'chain-build', target] + (' : '.join(urls)).split(' ') +  (['--nowait'] if is_nowait else [])
         print('  '.join(command), flush=True)
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -4,10 +4,11 @@ import os
 import subprocess
 import re
 from contextlib import contextmanager
-from uuid import uuid4
+from datetime import datetime, timedelta
 
 from specfile import Specfile
 
+TIME_FORMAT = '%Y-%m-%d-%H-%M-%S'
 
 @contextmanager
 def cd(dir):
@@ -59,12 +60,24 @@ def local_branch(branch):
     finally:
         subprocess.check_call(['git', 'checkout', prev_branch])
         subprocess.check_call(['git', 'branch', '-D', branch])
-        subprocess.check_call(['git', 'push', '--delete', 'origin', branch])
+
+def is_old_branch(b):
+    branch_time = datetime.strptime(b.split('/')[-1], TIME_FORMAT)
+    return branch_time < datetime.now() - timedelta(hours=3)
 
 def push_bumped_release(git_repo, test_build_id):
-    uuid = uuid4()
-    branch = f'koji/test/{uuid}'
+    t = datetime.now().strftime(TIME_FORMAT)
+    branch = f'koji/test/{test_build_id}/{t}'
     with cd(git_repo), local_branch(branch):
+        # clean up the old branches
+        subprocess.check_call(['git', 'fetch'])
+        remote_branches = subprocess.check_output(['git', 'branch', '-rl', 'origin/koji/test/*/*']).decode().splitlines()
+        remote_branches = [b.strip()[len('origin/'):] for b in remote_branches]
+        old_branches = [b for b in remote_branches if is_old_branch(b)]
+        for b in old_branches:
+            print(f"removing outdated remote branch {b}", flush=True)
+            subprocess.check_call(['git', 'push', '--delete', 'origin', b])
+        # locate the specfile
         spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec']).decode().splitlines()
         assert len(spec_paths) == 1
         spec_path = spec_paths[0]
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -4,6 +4,9 @@ import os
 import subprocess
 import re
 from contextlib import contextmanager
+from uuid import uuid4
+
+from specfile import Specfile
 
 
 @contextmanager
@@ -47,6 +50,32 @@ def koji_url(remote, hash):
         raise Exception("Unrecognized remote URL")
     return remote + "?#" + hash
 
+@contextmanager
+def local_branch(branch):
+    prev_branch = subprocess.check_output(['git', 'branch', '--show-current']).strip()
+    subprocess.check_call(['git', 'checkout', '-b', branch])
+    try:
+        yield branch
+    finally:
+        subprocess.check_call(['git', 'checkout', prev_branch])
+        subprocess.check_call(['git', 'branch', '-D', branch])
+        subprocess.check_call(['git', 'push', '--delete', 'origin', branch])
+
+def push_bumped_release(git_repo, test_build_id):
+    uuid = uuid4()
+    branch = f'koji/test/{uuid}'
+    with cd(git_repo), local_branch(branch):
+        spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec']).decode().splitlines()
+        assert len(spec_paths) == 1
+        spec_path = spec_paths[0]
+        with Specfile(spec_path) as spec:
+            # TODO: check koji build to use another final number when needed
+            spec.release = f'{spec.release}.0.{test_build_id}.1'
+        subprocess.check_call(['git', 'commit', '-m', "bump release for test build", spec_path])
+        subprocess.check_call(['git', 'push', 'origin', branch])
+        commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode().strip()
+        return commit
+
 def main():
     parser = argparse.ArgumentParser(description='Build a package or chain-build several from local git repos for RPM sources')
     parser.add_argument('target', help='Koji target for the build')
@@ -55,12 +84,14 @@ def main():
                              'a chained build will be started in the order of the arguments')
     parser.add_argument('--scratch', action="store_true", help='Perform scratch build')
     parser.add_argument('--nowait', action="store_true", help='Do not wait for the build to end')
+    parser.add_argument('--test-build', metavar="ID", help='Run a test build. The provided ID will be used to build a unique release tag.')
     args = parser.parse_args()
 
     target = args.target
     git_repos = [os.path.abspath(check_dir(d)) for d in args.git_repos]
     is_scratch = args.scratch
     is_nowait = args.nowait
+    test_build = args.test_build
 
     if len(git_repos) > 1 and is_scratch:
         parser.error("--scratch is not compatible with chained builds.")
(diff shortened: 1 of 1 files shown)


scripts/koji/koji_build.py
```diff
@@ -111,9 +111,11 @@ def main():
         urls = []
         for d in git_repos:
             remote, hash = get_repo_and_commit_info(d)
+            if test_build:
+                hash = push_bumped_release(d, test_build)
             urls.append(koji_url(remote, hash))
         command = ['koji', 'chain-build', target] + (' : '.join(urls)).split(' ') +  (['--nowait'] if is_nowait else [])
-        subprocess.check_call(['echo'] + command)
+        print('  '.join(command), flush=True)
```
ydirson, line 141: same as previous commit, all such changes should be grouped in a separate commit

scripts/koji/koji_build.py
```diff
@@ -50,6 +50,30 @@ def koji_url(remote, hash):
...
+        yield branch
+    finally:
+        subprocess.check_call(['git', 'checkout', prev_branch])
+        subprocess.check_call(['git', 'branch', '-D', branch])
+        subprocess.check_call(['git', 'push', '--delete', 'origin', branch])
+
+@contextmanager
+def push_bumped_release(git_repo, trigram):
+    uuid = uuid4()
+    branch = f'koji/test/{uuid}'
```
ydirson: maybe use the trigram in branch name too?

scripts/koji/koji_build.py
```diff
@@ -59,12 +60,23 @@ def local_branch(branch):
...
+
+def is_old_branch(b):
+    branch_time = datetime.strptime(b.split('/')[-1], TIME_FORMAT)
+    return branch_time < datetime.now() - timedelta(hours=3)
 
 def push_bumped_release(git_repo, trigram):
-    uuid = uuid4()
-    branch = f'koji/test/{uuid}'
+    t = datetime.now().strftime(TIME_FORMAT)
+    branch = f'koji/test/{trigram}/{t}'
```
ydirson: that part could indeed be squashed into original commit

scripts/koji/koji_build.py
```diff
@@ -59,12 +60,23 @@ def local_branch(branch):
...
     with cd(git_repo), local_branch(branch):
+        # clean up the old branches
+        subprocess.check_call(['git', 'fetch'])
+        remote_branches = subprocess.check_output(['git', 'branch', '-rl', 'origin/koji/test/*/*']).decode('utf8').splitlines()
+        remote_branches = [b.strip()[len('origin/'):] for b in remote_branches]
+        old_branches = [b for b in remote_branches if is_old_branch(b)]
+        for b in old_branches:
+            print(f"removing outdated remote branch {b}", flush=True)
+            subprocess.check_call(['git', 'push', '--delete', 'origin', b])
```
ydirson: I'm dubious on the feature:
- either we allow to run several such test builds concurrently (which I think is a good idea, or at least I don't really see a reason not to), and then we don't want to destroy any branches
- or we don't allow it, and we might as well just use a unique branch name for all such tests

Further more, there should be very little space for such branches to linger, with context managers doing their job - only killing the script in a way python cannot recover should leave any such trace behind.

scripts/koji/koji_build.py
```diff
@@ -79,8 +79,15 @@ def push_bumped_release(git_repo, trigram):
             subprocess.check_call(['git', 'push', '--delete', 'origin', b])
         spec_path = glob('SPECS/*.spec')[0]
         with Specfile(spec_path) as spec:
-            # TODO: check koji build to use another final number when needed
-            spec.release = f'{spec.release}.0.{trigram}.1'
+            # find the next build number
+            package = spec_path.split('/')[-1][:-len('.spec')]
```
ydirson, line 85: isn't it more safely extracted from the spec object?
(if not, at least use `os.path.basename`)

scripts/koji/koji_build.py
```diff
@@ -79,8 +79,15 @@ def push_bumped_release(git_repo, trigram):
...
+            builds = subprocess.check_output(['koji', 'list-builds', '--package', package]).decode('utf8').splitlines()[2:]
+            base_name = f'{package}-{spec.version}-{spec.release}.0.{trigram}.'
```
ydirson, line 87: "name" (here and in the commit message) is misleading

scripts/koji/koji_build.py
```diff
@@ -79,8 +79,15 @@ def push_bumped_release(git_repo, trigram):
...
+            base_name_re = re.escape(re.sub('%{.+}', "@@@", base_name)).replace('@@@', '.*') + r'(\d+)'
+            build_matchs = [re.match(base_name_re, b) for b in builds]
```
ydirson, line 89: that looks a bit complicated, what about just filtering on `startswith(basename)` and stripping the first `len(basename)` chars?

scripts/koji/koji_build.py
stormi: You are encouraged to use a branch, we just won't pollute it for the build.

scripts/koji/koji_build.py
```diff
@@ -3,35 +3,42 @@
...
+from contextlib import contextmanager, nullcontext
+from glob import glob
+from uuid import uuid4
+
+from specfile import Specfile
+
+
+@contextmanager
+def cd(dir, internal=False, makedirs=False):
+    """Change to a directory temporarily. To be used in a with statement"""
```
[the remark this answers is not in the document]
glehmann: I introduced it to work with multiple git commands and deal with potential failures. I've put it in its own commit to keep the next commits as atomic as possible.
Indeed the commit message could have been made clearer!
I never used the `git` module, but I could give it a try

scripts/koji/koji_build.py
```diff
@@ -50,6 +50,30 @@ def koji_url(remote, hash):
...
 
+@contextmanager
+def local_branch(branch):
+    prev_branch = subprocess.check_output(['git', 'branch', '--show-current']).strip()
+    subprocess.check_call(['git', 'checkout', '-b', branch])
+    try:
+        yield branch
+    finally:
+        subprocess.check_call(['git', 'checkout', prev_branch])
+        subprocess.check_call(['git', 'branch', '-D', branch])
```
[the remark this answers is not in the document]
glehmann: good point, didn't thought about that!

scripts/koji/koji_build.py
```diff
@@ -50,6 +50,30 @@ def koji_url(remote, hash):
...
+        subprocess.check_call(['git', 'push', '--delete', 'origin', branch])
```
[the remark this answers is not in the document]
glehmann: yes, it disappears later in the commits

scripts/koji/koji_build.py
```diff
@@ -50,6 +50,30 @@ def koji_url(remote, hash):
...
+
+@contextmanager
+def push_bumped_release(git_repo, trigram):
+    uuid = uuid4()
+    branch = f'koji/test/{uuid}'
+    with cd(git_repo), local_branch(branch):
+        spec_path = glob('SPECS/*.spec')[0]
+        with Specfile(spec_path) as spec:
+            # TODO: check koji build to use another final number when needed
+            spec.release = f'{spec.release}.0.{trigram}.1'
```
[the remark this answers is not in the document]
glehmann: not when used as a context manager!

scripts/koji/koji_build.py
glehmann: we have to keep the remote branch if we want to run with `--nowait`. The context manager won't do the job in that case, because it's gone long before the build is done (or even started)

scripts/koji/koji_build.py
glehmann: none that I've found unfortunately, but I'm open to suggestions :-)
`basename` is a good idea though

scripts/koji/koji_build.py
glehmann: base_name might include some macros, so it wouldn't match. So this is a (not so) complicated way to match without precisely knowing to what value the macro is expanded

scripts/koji/koji_build.py
```diff
@@ -3,35 +3,42 @@
...
+
+
+@contextmanager
+def cd(dir, internal=False, makedirs=False):
+    """Change to a directory temporarily. To be used in a with statement"""
+    prevdir = os.getcwd()
+    try:
+        yield os.path.realpath(dir)
+    finally:
+        os.chdir(prevdir)
```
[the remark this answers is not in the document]
glehmann: I'm not sure I understand that comment. What do you mean with visible?

scripts/koji/koji_build.py
glehmann: I did that latter in the commits

scripts/koji/koji_build.py
```diff
@@ -73,11 +99,13 @@ def main():
...
-        url = koji_url(remote, hash)
-        command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
-        subprocess.check_call(['echo'] + command)
-        subprocess.check_call(command)
+        with push_bumped_release(git_repos[0], test_build) if test_build else nullcontext():
+            remote, hash = get_repo_and_commit_info(git_repos[0])
+            url = koji_url(remote, hash)
+            command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
+            print('  '.join(command))
+            command = ['false']
```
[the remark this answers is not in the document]
glehmann: yes

scripts/koji/koji_build.py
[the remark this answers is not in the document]
glehmann: I tried it, but it makes git a lot more verbose when working on the detached head, so I kept the temporary local branch

```patch
diff --git a/scripts/koji/koji_build.py b/scripts/koji/koji_build.py
index ee5debf..74cb951 100755
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -53,12 +53,12 @@ def koji_url(remote, hash):
 @contextmanager
 def local_branch(branch):
     prev_branch = subprocess.check_output(['git', 'branch', '--show-current']).strip()
-    subprocess.check_call(['git', 'checkout', '-b', branch])
+    commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf8').strip()
+    subprocess.check_call(['git', 'checkout', commit])
     try:
         yield branch
     finally:
         subprocess.check_call(['git', 'checkout', prev_branch])
-        subprocess.check_call(['git', 'branch', '-D', branch])
         subprocess.check_call(['git', 'push', '--delete', 'origin', branch])
 
 @contextmanager
@@ -73,7 +73,7 @@ def push_bumped_release(git_repo, trigram):
             # TODO: check koji build to use another final number when needed
             spec.release = f'{spec.release}.0.{trigram}.1'
         subprocess.check_call(['git', 'commit', '-m', "bump release for test build", spec_path])
-        subprocess.check_call(['git', 'push', 'origin', branch])
+        subprocess.check_call(['git', 'push', 'origin', f'HEAD:refs/heads/{branch}'])
         yield branch
 
 def main():
```

scripts/koji/koji_build.py
```diff
@@ -3,35 +3,42 @@
 import os
 import subprocess
 import re
+from contextlib import contextmanager, nullcontext
+from glob import glob
+from uuid import uuid4
+
+from specfile import Specfile
```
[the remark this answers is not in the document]
glehmann: mmm, they seems to be where they belong. Could you tell why they should be in the next commit?

scripts/koji/koji_build.py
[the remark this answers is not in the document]
ydirson: Can be done quite unverbosely withsomething like
```
(env-mypy) user@work-vates:scripts (ydi/repo-status =)$ git checkout --quiet HEAD^{commit} 
(env-mypy) user@work-vates:scripts ((3d5f620...))$ 
```

scripts/koji/koji_build.py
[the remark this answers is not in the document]
ydirson: but shouldn't it be called by `push_bumped_release` anyway?

scripts/koji/koji_build.py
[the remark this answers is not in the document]
ydirson: I mean that even after looking a few times at the code, I don't see the call to `os.chdir(dir)` which the `finally` clause is meant to revert.  I would have expected it inside this contextmanager, but really can't find it anywhere else either.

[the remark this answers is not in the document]
glehmann: oh! that's clearly a problem :+1:

scripts/koji/koji_build.py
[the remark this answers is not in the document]
ydirson: Because in the commit introducing them, only `contextmanager` gets actually used.  All others are only needed by the next commits.  With recent changes `glob` is even never used.

scripts/koji/koji_build.py
```diff
@@ -73,11 +99,13 @@ def main():
...
-        remote, hash = get_repo_and_commit_info(git_repos[0])
-        url = koji_url(remote, hash)
-        command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
-        subprocess.check_call(['echo'] + command)
-        subprocess.check_call(command)
+        with push_bumped_release(git_repos[0], test_build) if test_build else nullcontext():
+            remote, hash = get_repo_and_commit_info(git_repos[0])
+            url = koji_url(remote, hash)
+            command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
+            print('  '.join(command))
```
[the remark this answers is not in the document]
ydirson: Such changes when in an atomic commit are really easy to review and understand, whereas intermixed with other changes they indeed make the other commit harder to read, since additional work is requested from the reader to filter out this stuff.
My position is that there is a balance to be found when complicated changes indeed interact, but here it is not the case.

scripts/koji/koji_build.py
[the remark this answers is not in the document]
ydirson: but it is still there ;)

scripts/koji/koji_build.py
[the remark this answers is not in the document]
glehmann: yes, I'm not sure how I read the diff in github, but I can see what's wrong without problem with `git diff`
should be fixed now

scripts/koji/koji_build.py
ydirson: Yes, but in a commit that already does 2 other things (switch from uuid to date, and branch cleanup).  It would be better to fold the switch to `f'koji/test/{trigram}/{t}'` completely in the original commit, so `uuid` also never gets introduced.  That way this other commit will have the single purpose of branch cleanup.

scripts/koji/koji_build.py
ydirson: oh, that one is now part of the previous thread

scripts/koji/koji_build.py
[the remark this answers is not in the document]
glehmann: yes it should, but I changed the logic in a later commit to stop removing the remote branch in the context manager, so I didn't bother fixing it here. Do you think that's a problem?

scripts/koji/koji_build.py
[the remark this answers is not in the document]
glehmann: Ok, I don't find it difficult to read personally, but I hear your point, and I'll try to avoid that in the future :+1:

scripts/koji/koji_build.py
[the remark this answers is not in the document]
glehmann: really?

scripts/koji/koji_build.py
```diff
@@ -99,13 +99,14 @@ def main():
...
-            command = ['false']
-            subprocess.check_call(command)
+        remote, hash = get_repo_and_commit_info(git_repos[0])
+        if test_build:
+            hash = push_bumped_release(git_repos[0], test_build)
+        url = koji_url(remote, hash)
+        command = ['koji', 'build'] + (['--scratch'] if is_scratch else []) + [target, url] + (['--nowait'] if is_nowait else [])
+        print('  '.join(command), flush=True)
+        # command = ['false']
+        subprocess.check_call(command)
```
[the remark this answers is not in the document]
glehmann: indeed it wasn't working when deleting the remote branch, because koji is not able to checkout on a commit that is not part of a branch.
But it didn't work with `--nowait`  either so I switch to cleaning the remote branch at the next run, to make `--nowait` compatible with `--test-build`

scripts/koji/koji_build.py
glehmann: I've used `Path(spec_path).stem` — works great and avoid also calling `os.path.splitext`

scripts/koji/koji_build.py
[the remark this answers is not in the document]
stormi: > sure that would be better in its own commit, but for some small changes like this one, I find acceptable to include it in the another commit and avoid the burden to make another commit later. It's about finding a good balance between commit history and coding flow :-)

I can agree with this view in general. No opinion about the current case :)

scripts/koji/koji_build.py
```diff
@@ -43,6 +52,53 @@ def koji_url(remote, hash):
...
+    t = datetime.now().strftime(TIME_FORMAT)
+    branch = f'koji/test/{trigram}/{t}'
+    with cd(git_repo), local_branch(branch):
+        spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec']).decode('utf8').splitlines()
+        assert len(spec_paths) == 1
+        spec_path = spec_paths[0]
+        with Specfile(spec_path) as spec:
+            # find the next build number
+            package = Path(spec_path).stem
+            builds = subprocess.check_output(['koji', 'list-builds', '--package', package]).decode('utf8').splitlines()[2:]
```
stormi, line 89: If one wants to send a similar build to both their test branches on 8.2 and 8.3, as I understand the current code, one will end in `.1.xcpng8.2` and the other will end in `.2.xcpng8.3` when they would have expected `.1.xcpng8.2` and `.1.xcpng8.3` instead.

The problem is, we can't extract from the spec file the value of `%dist`. It depends on macros defined in the build root. However, a possible heuristic is that per convention, target `vX.Y-u-user1` is a build for the `X.Y` release of XCP-ng, so `%{?dist}` or `%dist` will evaluate to `xcpngX.Y`.

scripts/koji/koji_build.py
```diff
@@ -43,6 +52,53 @@ def koji_url(remote, hash):
...
+            base_nvr = f'{package}-{spec.version}-{spec.release}.0.{trigram}.'
+            # use a regex to match %{macro} without actually expanding the macros
+            base_nvr_re = re.escape(re.sub('%{.+}', "@@@", base_nvr)).replace('@@@', '.*') + r'(\d+)'
+            build_matchs = [re.match(base_nvr_re, b) for b in builds]
```
stormi, line 93: Nitpicking: `build_matches` looks more consistent with english spelling rules.

```suggestion
            build_matches = [re.match(base_nvr_re, b) for b in builds]
```

scripts/koji/koji_build.py
```diff
@@ -43,6 +52,53 @@ def koji_url(remote, hash):
...
+            build_ids = [int(m.group(1)) for m in build_matchs if m]
+            next_build_id = sorted(build_ids)[-1] + 1 if build_ids else 1
+            spec.release = f'{spec.release}.0.{trigram}.{next_build_id}'
```
stormi, line 96: What does `%{?dist}` become? Lost?

scripts/koji/koji_build.py
```diff
@@ -43,6 +52,53 @@ def koji_url(remote, hash):
         raise Exception("Unrecognized remote URL")
     return remote + "?#" + hash
 
+@contextmanager
+def local_branch(branch):
+    prev_branch = subprocess.check_output(['git', 'branch', '--show-current']).strip()
+    commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf8').strip()
```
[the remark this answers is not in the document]
glehmann: Well, that's an old habit, but I'm not sure where it comes. Maybe from python 2?
I'll change it.

scripts/koji/koji_build.py
glehmann: yes, the sub-release number (not sure how to call it!) would increase for all the targets.
But I'm not sure that's a problem — having a number incremented with the build seems good enough. It's just a test build, there is no real impact if we skip a number

scripts/koji/koji_build.py
glehmann: `specfile` doesn't include it in the `release`. I'm not sure where it puts it in memory, but it's still in the modified specfile.

scripts/koji/koji_build.py
stormi: But it makes no sense to skip one.

glehmann: I usually prefer to keep things simple, and it seemed good enough to me that way. Simple and good enough is usually a good combination :-)
But I agree that your proposed change is semantically better. It's implemented.

stormi: I agree, it's not easy to draw the line, but thanks for implementing it!

scripts/koji/koji_build.py
```diff
@@ -1,37 +1,47 @@
...
 import argparse
+import logging
 import os
 import subprocess
 import re
+from contextlib import contextmanager
+from datetime import datetime, timedelta
+from pathlib import Path
+
+from specfile import Specfile
```
stormi, line 11: I was testing but forgot to add the extra dependency. And this will happen to others, as we don't package it at the moment.

```
~/Vates/git/rpms/dummypkg [stormi-test L ✔] $ koji_build.py --help
Traceback (most recent call last):
  File "/home/samuel/Vates/git/xcp/scripts/koji/koji_build.py", line 11, in <module>
    from specfile import Specfile
ModuleNotFoundError: No module named 'specfile'
```

I suggest catching the error and turning it into a nice message on the console, for a smoother transition of existing users of this script.

stormi (approved):
Outside my suggestion regarding import errors, looks good to me.

I used it to do two test builds.

scripts/koji/koji_build.py
```diff
@@ -8,7 +8,11 @@
 from datetime import datetime, timedelta
```
stormi, line 8: Looks like this last commit is missing the DCO

PR #683, opened 2025-03-07
Fix several issues in import_srpm

ydirson:


scripts/import_srpm.py
```diff
@@ -1,20 +1,34 @@
 #!/usr/bin/env python3
 import argparse
+import logging
 import os
 import subprocess
 
+def call_process(args):
```
glehmann, line 7: `call` is probably clear enough in that context, and pleasantly shorter

glehmann (approved):

stormi (approved):
The internal wiki will need updating at the same time as this merge.

scripts/import_srpm.py
ydirson: Hm, I'm a bit wary of using generic names, especially when it's so cheap to be more specific

ydirson:
Added another small fix still in the "fix import" scope, adjusting PR name to encompass it

glehmann (approved):

stormi (approved):
Approved, but please update the wiki *just before* merging this.

This will impact current habits, notably those of @gduperrey.

gduperrey:
> Approved, but please update the wiki _just before_ merging this.
> 
> This will impact current habits, notably those of @gduperrey.

Yes indeed. If this changes the way it's used, the documentation citing this script on the wiki must be updated.

There's at least one for hotfix management, but there may be others.

PR #688, opened 2025-04-07
ignore helix directory

glehmann:
so I can deal with repository specific configurations without committing the configuration


glehmann:
just remembered that I could use `.git/info/exclude`

PR #695, opened 2025-04-23
code owners must be on the same line if they use the same pattern

glehmann:
with multiple matching patterns, only the last one is used


stormi:
I don't understand this "patterns must be on the same line" thing. Can you explain?

glehmann:
> If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners are not on the same line, the pattern matches only the last mentioned code owner.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

stormi (approved):

PR #699, opened 2025-04-25
use git ls-remote to search the old branches to remove

glehmann:
git branch can't be used with these flags with git 1.8 that we have in xcp-ng-build-env


scripts/pkg_in_pipe/report.html
stormi, line 1: Was this file included on purpose?

glehmann: Not at all!

stormi (approved):

PR #706, opened 2025-05-12
use less vertical space

glehmann:


stormi:
Can you provide a screenshot?

glehmann:
> Can you provide a screenshot?

sure, I've updated the PR description with the screenshots

scripts/pkg_in_pipe/pkg_in_pipe.py
```diff
@@ -136,20 +136,20 @@ def print_table_line(out, build, link, issues, built_by, prs: list[PullRequest],
             <th scope="row" class="px-6 py-4 font-medium text-gray-900 whitespace-nowrap dark:text-white">
```
stormi, line 136: Why not here too?

glehmann: That's an overlook. The result is even more dense with it, perhaps a bit too much to my taste

stormi: We can adjust later if that's too much, then.

stormi (approved):

PR #709, opened 2025-05-12
alert on problems with warning signs

glehmann:
* when the build is not found in any of the plane card
* when the PR is not found in any of the plane card
* when the package has not maintainer

image

### DIFF
--- a/scripts/pkg_in_pipe/pkg_in_pipe.py
+++ b/scripts/pkg_in_pipe/pkg_in_pipe.py
@@ -89,11 +89,11 @@ def print_table_header(out, tag):
                         <th scope="col" class="px-6 py-2 w-[25.5%]">
                             Build
                         </th>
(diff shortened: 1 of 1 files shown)


glehmann:
>     * when the PR is not found in any of the plane card

I'm re-reading that, and I'm not sure that's the right check. Do we have to copy/paste the build into each related card?
Currently, this check ensures that the build appears in at least one of the cards, but not all of them.

stormi:
> > ```
> > * when the PR is not found in any of the plane card
> > ```
> 
> I'm re-reading that, and I'm not sure that's the right check. Do we have to copy/paste the build into each related card? Currently, this check ensures that the build appears in at least one of the cards, but not all of them.

I'm unsure myself. Let's start with the way you defined the check and see if there are future cases where we find it not sufficient.

stormi (approved):

PR #710, opened 2025-05-13
koji: explicitly require the specfile python package

glehmann:


Nambrok (approved):
Adding a README, other informations should be added to it eventually.

last-genius (approved):

PR #711, opened 2025-05-19
Fix too many PRs listed when a more recent build exist in a tag with …

glehmann:
…higher priority

that can happen when a security fix is fasttracked


stormi (approved):
Looks good, but would look even better with a commit title shorter than 70 chars.

PR #716, opened 2025-05-22
Avoid the blank page during the report generation

glehmann:
by writing the whole report at once in the output file.


stormi (approved):
Fast-tracking as I need that fix right now :)

PR #717, opened 2025-05-22
also search for the previous build in the current tag

glehmann:


stormi:
Another fast-track to get the fix fast.

PR #719, opened 2025-05-22
Sort the builds based on the tagging timestamp

glehmann:


scripts/pkg_in_pipe/pkg_in_pipe.py
```diff
@@ -315,9 +315,13 @@ def find_pull_requests(gh, repo, start_sha, end_sha):
             session = koji.ClientSession('https://kojihub.xcp-ng.org', config)
             session.ssl_login(config['cert'], None, config['serverca'])
             for tag in tags:
+                tag_history = dict(
+                    (tl['build_id'], tl['create_ts']) for tl in session.queryHistory(tag=tag)['tag_listing']
```
stormi, line 319: This can be a slow request, if I trust the delays on CLI when I query the history. Isn't there any filtering option that can speed things up? We don't need the full history, we only need the history relative to the builds that are currently in the tag. So I believe, if the API allows it, that we could filter either on the list of builds in the tag, or, if the API allows it and if it makes sense, an equivalent of `--active` on the CLI. I'm not 100% sure what it means, but it seems that it limited the history to pertinent events, including events that concern builds currently in the tag, and it was faster.

glehmann: The query is already filtering on the tag.
I tried to add `active=True` in the parameters and its significantly faster (~5s) :+1:

stormi: Yes, when I meant slow request, that was even when filtering on a given tag. Good for `active=True`.

stormi (approved):

PR #728, opened 2025-06-18
Ensure to build the last commit of the expected branch for protected targets

glehmann:
We don't check that the commit is in the remote branch, in order to avoid building from a commit in a PR that has been merged but is not the merged commit.

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -16,6 +16,16 @@ except ImportError:
 
 TIME_FORMAT = '%Y-%m-%d-%H-%M-%S'
 
+# target -> required branch
+PROTECTED_TARGETS = {
+    "v8.2-ci": "8.2",
+    "v8.2-fasttrack": "8.2",
+    "v8.2-incoming": "8.2",
+    "v8.3-ci": "master",
+    "v8.3-fasttrack": "master",
+    "v8.3-incoming": "master",
+}
+
 @contextmanager
 def cd(dir):
     """Change to a directory temporarily. To be used in a with statement."""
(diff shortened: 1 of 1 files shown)


scripts/iso-remaster/samples/patch-installimg.sh
```diff
@@ -5,11 +5,11 @@ set -e
 INSTALLIMG="$1"
 
 ## Include a locally-modified version of the installer for testing
-#HOSTINSTALLER=$HOME/src/xs/host-installer
+HOSTINSTALLER=$HOME/src/vates/host-installer
 ## - if 8.2: copies a few files too much until we can "make install", but that's harmless
 #cp -rv "$HOSTINSTALLER"/* "$INSTALLIMG/opt/xensource/installer/"
 ## - if 8.3
-#make -C "$HOSTINSTALLER" DESTDIR="$INSTALLIMG" XS_MPATH_CONF="$HOME/src/xapi/sm/multipath/multipath.conf"
+make -C "$HOSTINSTALLER" DESTDIR="$INSTALLIMG" XS_MPATH_CONF="/home/glehmann/src/vates/sm/multipath/multipath.conf"
```
stormi: Doesn't look like it belongs to this PR

scripts/koji/koji_build.py
```diff
@@ -36,10 +46,17 @@ def check_git_repo(dirpath):
...
+def check_commit_is_available_remotely(dirpath, hash, target):
     with cd(dirpath):
         if not subprocess.check_output(['git', 'branch', '-r', '--contains', hash]):
             raise Exception("The current commit is not available in the remote repository")
+        expected_branch = PROTECTED_TARGETS.get(target)
+        if (
+            expected_branch is not None
+            and not is_remote_branch_commit(dirpath, hash, expected_branch)
+        ):
+            raise Exception(f"The current commit is not the last commit in the remote branch {expected_branch}.\n"
```
stormi: I have doubts about the heuristics. We could try with this one, but I think there must be an option to force-build. Especially with `fasttrack`, where there are cases when we may have to build from a branch because master has too many changes and we want to push an update out with just a tiny XSA fix.

stormi (approved):

scripts/koji/koji_build.py
```diff
@@ -36,10 +46,23 @@ def check_git_repo(dirpath):
...
         return subprocess.run(['git', 'diff-index', '--quiet', 'HEAD', '--']).returncode == 0
 
-def check_commit_is_available_remotely(dirpath, hash):
+def check_commit_is_available_remotely(dirpath, hash, target, warn):
     with cd(dirpath):
-        if not subprocess.check_output(['git', 'branch', '-r', '--contains', hash]):
-            raise Exception("The current commit is not available in the remote repository")
+        try:
+            if not subprocess.check_output(['git', 'branch', '-r', '--contains', hash]):
+                raise Exception("The current commit is not available in the remote repository")
```
ydirson: Do we want this to be part of the `try` block?  We may not want `--force` to override that check, since the build is indeed sure to fail, right?

stormi:
Ping @ydirson. We just had a build made from the PR branch instead of master :)

PR #732, opened 2025-06-26
convert plane data to grist document

glehmann:


stormi:
I'd like to come back to this sometime, but for now it's probably better to convert it to draft.

PR #733, opened 2025-07-04
python project configuration for koji scripts

glehmann:
We can now `uv sync` in the dir, and `export PATH=$PWD/.venv/bin:$PATH` to use `koji-build` with its required dependency.

The package should also be publishable, allowing, in the future, to `pipx install koji-utils` to have `koji-build` and the other tools available in the shell.

Note that I had to move `koji_build.py` to a module directory. The root `koji_build.py` now prints a warning saying it's been moved, but we could also *not* print any warning.


stormi:
Does this incur a change for current users of the `koji_build.py` script?

scripts/koji/koji_build.py
```diff
@@ -1,204 +1,9 @@
...
 
-try:
-    from specfile import Specfile
-except ImportError:
-    print("error: specfile module can't be imported. Please install it with 'pip install --user specfile'.")
-    exit(1)
+import sys
 
-TIME_FORMAT = '%Y-%m-%d-%H-%M-%S'
+from koji_utils.koji_build import main
```
ydirson, line 5: Could have a comment (and possibly a note printed) telling it is a now deprecated compatibility script?

Alternatively, just using symlinks would be simpler and likely more clear on which script version gets launched (the one in the same repo or the one previously installed).  It would also allow to relocate all the scripts into the new package first, and then add the entrypoints.

scripts/koji/pyproject.toml
```diff
@@ -8,6 +8,16 @@ dependencies = [
   "specfile",
 ]
 
+[build-system]
+requires = ["uv_build>=0.7.19,<0.8.0"]
+build-backend = "uv_build"
```
ydirson: I'm curious, what does this build backend brings over `setup_tools`?  This kind of design choice is useful to note in the commit message.

ydirson (approved):
The info in the commit message would be useful in a README, and maybe `uvx` can be suggested for install?

scripts/koji/pyproject.toml
glehmann: I just used it to match `uv`. @psafont used `setuptools` in https://github.com/xcp-ng/python-project-template/pull/1 and I'm also fine with it. I'll update this PR to use `setuptools`.

glehmann:
> Does this incur a change for current users of the `koji_build.py` script?

Currently it only display a warning about the file location that has been changed. But that's not mandatory.

scripts/koji/pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[project]
+name = "koji-utils"
+version = "0.1.0"
+description = "Koji utils"
+readme = "README.md"
+requires-python = "~=3.11"
```
psafont, line 5: This version is ambiguous and makes uv throw warnings, use ">=3.11, <4" instead

scripts/koji/.python-version
```diff
@@ -0,0 +1 @@
+3.11.11
```
psafont, line 1: What's the point of this file? This version can't be enforced (my system doesn't provide it)

scripts/koji/setup.cfg
```diff
@@ -0,0 +1,7 @@
+[flake8]
```
psafont, line 1: Can't this be part of pyproject.toml?

scripts/koji/.python-version
ydirson: `uv` uses it to pin the python version

scripts/koji/pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[project]
```
psafont, line 1: I don't see much point of this, if it's not going to install any script, In any case I'd rather only have a single pyproject.toml file at the root, it can be limited to only detect a single package from the koji directory:

```
[project]
name = "xcp"
requires-python = ">= 3.11, <4"
dynamic = ["version"]
readme = "README.md"
dependencies = ["specfile >= 0.34.0"]

[project.scripts]
"koji-xcp" = "koji.koji_build:main"

[build-system]
requires = ["setuptools >= 77.0.3", "setuptools-scm>=8"]
build-backend = "setuptools.build_meta"

[tool.setuptools]
package-dir = {"" = "scripts"}
packages = ["koji"]

[tool.setuptools_scm]

```

scripts/koji/koji_build.py
glehmann: > Could have a comment (and possibly a note printed) telling it is a now deprecated compatibility script?

like the one 2 lines below?

> Alternatively, just using symlinks would be simpler and likely more clear on which script version gets launched (the one in the same repo or the one previously installed). It would also allow to relocate all the scripts into the new package first, and then add the entrypoints.

I've also considered that, but printing a warning is simpler this way

scripts/koji/pyproject.toml
psafont: setuptools allows to use setuptools_scm, which automagivally infers the version using git, which I very much prefer to manually tinkering with file, or doing replacements of placeholder that I've seen hand-rolled elsewhere

scripts/koji/.python-version
glehmann: > What's the point of this file? This version can't be enforced (my system doesn't provide it.

`uv` should install that exact version for you. That's not the case?

scripts/koji/pyproject.toml
glehmann: In my opinion we should split this project in their own repository.
That would be easier to properly tag versions, make releases, run the CI,…

scripts/koji/koji_utils/koji_build.py
```diff
@@ -0,0 +1,204 @@
...
+            if test_build or pre_build:
+                hash = push_bumped_release(d, target, test_build, pre_build)
+            else:
+                check_commit_is_available_remotely(d, hash)
+            urls.append(koji_url(remote, hash))
+        command = ['koji', 'chain-build', target] + (' : '.join(urls)).split(' ') + (['--nowait'] if is_nowait else [])
+        print('  '.join(command), flush=True)
+        subprocess.check_call(command)
+
+if __name__ == "__main__":
```
psafont, line 250: Why is all this code being moved? I was able to expose the cli script without having to move any: https://github.com/xcp-ng/xcp/commit/c4993252f39f2e11813ef8bb35223722b5f292ae

scripts/koji/setup.cfg
glehmann: flake8 doesn't support `pyproject.toml` natively.
I think there is a plugin for that though. Do you think that would be better?

scripts/koji/.python-version
psafont: I install tools using uv tool, I've never seen it installing python version.

scripts/koji/pyproject.toml
psafont: Absolutely

scripts/koji/setup.cfg
psafont: I think it's fine, I was just curious

scripts/koji/.python-version
ydirson: Feels strange, as https://docs.astral.sh/uv/guides/tools/ says `uv tool run` is the same as `uvx`, and
> Tools are installed into temporary, isolated environments when using uvx.

scripts/koji/pyproject.toml
ydirson: If we go that way, maybe there is no point in keeping the `koji_build.py` compat script?

scripts/koji/koji_utils/koji_build.py
glehmann: I couldn't make it work without moving the code in its own module, but that was with a local `pyproject.toml`, not the global one, and with `uv_build` instead of `setuptools`

scripts/koji/.python-version
psafont: If the system python matches the version allowed in pyproject.toml, I don't see why it would be strange to use the system interpreter

psafont (approved):

ydirson (approved):

stormi:
I removed myself from the reviewers. There are two approvals, looks like it can be rebased and then merged.

stormi:
@glehmann This PR stalled for no apparent reason.

PR #739, opened 2025-07-29
Merge packages_provenance.csv and package_status.csv in packages.json

glehmann:


scripts/rpm_owners/packages.json
```diff
@@ -1,2890 +1,4924 @@
 {
     "GConf2": {
+        "added_by": "?",
```
ydirson: This looks like an automated process, we may want to document how it was done?

glehmann: It was processed in an interactive Python session. I'm afraid there isn't much to document, apart from the data sources, which are in the commit description.

ydirson (approved):
nothing much to review then :)

stormi (approved):

PR #740, opened 2025-08-04
import_srpm: allow to pass a url as source file

glehmann:
This way we don't even have to manually download the srpm.

### DIFF
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -159,7 +159,7 @@ def main():
         print(" committing...")
         has_changes = False
         try:
-            call_process(['git', 'diff-index', '--quiet',  'HEAD', '--'])
+            call_process(['git', 'diff-index', '--quiet', 'HEAD', '--'])
         except:
             has_changes = True
 
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -103,7 +103,6 @@ def main():
             call_process(['git', 'diff-index', '--quiet', 'HEAD', '--'])
             print("Working copy is clean.")
         except Exception:
-            raise
             parser.error("Git repository seems to have local modifications.")
 
         # check that there are no untracked files
@@ -123,8 +122,8 @@ def main():
         if os.path.isdir('SOURCES') and len(os.listdir('SOURCES')) > 0:
             call_process(['git', 'rm', 'SOURCES/*', '-r'])
         if os.path.isdir('SOURCES') and len(os.listdir('SOURCES')) > 0:
-            parser.error("Files remaining in SOURCES/ after removing the tracked ones. ")
-            parser.error("Delete them (including hidden files), reset --hard.")
+            parser.error("Files remaining in SOURCES/ after removing the tracked ones. \n"
+                         "Delete them (including hidden files), reset --hard.")
         os.mkdir('SOURCES')
 
         if os.path.isdir('SPECS'):
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -146,7 +146,8 @@ def main():
         for f in ['Citrix_Logo_Black.png', 'COPYING.CitrixCommercial']:
             if f in sources:
                 os.unlink(os.path.join('SOURCES', f))
-                open(os.path.join('SOURCES', "%s.deleted-by-XCP-ng.txt" % f), 'w').write(deletemsg)
+                with open(os.path.join('SOURCES', "%s.deleted-by-XCP-ng.txt" % f), 'w') as f:
+                    f.write(deletemsg)
                 deleted.append(f)
 
         if subprocess.call(['git', 'rev-parse', '--quiet', '--verify', args.branch]) != 0:
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -102,7 +102,7 @@ def main():
         try:
             call_process(['git', 'diff-index', '--quiet', 'HEAD', '--'])
             print("Working copy is clean.")
-        except:
+        except Exception:
             raise
             parser.error("Git repository seems to have local modifications.")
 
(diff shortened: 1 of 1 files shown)


scripts/import_srpm.py
```diff
@@ -28,12 +38,19 @@ def main():
                 }[args.verbose]
     logging.basicConfig(format='[%(levelname)s] %(message)s', level=loglevel)
 
+    source_rpm = args.source_rpm
+    if is_url(source_rpm):
+        # get the src.rpm locally, and continue with the actual file
+        local_filename = f'/tmp/{os.path.basename(source_rpm)}'
```
stormi: Should use a temporary directory known not to conflict with anything

stormi: And it could also self-destroy at the end, as python's `tempfile` (IIRC) allows?

stormi:
@glehmann ready for re-review?

stormi:
@glehmann you pushed after my last question but didn't answer nor ask for a re-review. Thus the PR stalled. What's the current status?

glehmann:
The changes were meant to be minimal, hence the lack of temp file removal.
I've updated the PR to properly remove the tempfile, at the cost of an indentation of most of the code

scripts/import_srpm.py
```diff
@@ -64,110 +75,120 @@ def main():
...
-        call_process(['git', 'tag', args.tag])
-
-    # push to remote
-    if args.push:
-        call_process(['git', 'push', '--set-upstream', 'origin', args.branch])
+    # handle URL source
+    source_rpm = args.source_rpm
+    is_remote_rpm = is_url(source_rpm)
+
+    with TemporaryDirectory() if is_remote_rpm else nullcontext() as temp_dir:
```
stormi, line 82: I believe that you could still keep the changes minimal as initially intended by avoiding the use of a context manager. You could use `atexit.register(temp_dir.cleanup)` to ensure it's cleaned up even in case of a raised exception. Unless I missed something obvious.

Not a blocker comment.

stormi (approved):

scripts/import_srpm.py
glehmann: I think it's cleaner that way, so I'd prefer to keep the context manager.
The diff is hard to read in github, but with better diff viewers, it's easy to see that it's mostly indentation changes :)

ydirson: To make such diffs more readable, a good way is to split things in different commits.  Here that could be "add a nullcontext ctxmanager" as a first step.

scripts/import_srpm.py
```diff
@@ -64,110 +75,120 @@ def main():
...
+        if not os.path.isdir(args.repository):
+            parser.error("Repository directory %s does not exist." % args.repository)
+        os.chdir(args.repository)
+
+        # check that the working copy is clean
+        try:
+            call_process(['git', 'diff-index', '--quiet', 'HEAD', '--'])
+            print("Working copy is clean.")
+        except Exception:
+            parser.error("Git repository seems to have local modifications.")
```
ydirson, line 106: That is an unrelated change that does not belong to this commit.  Likely a fix for a real bug, since the original code really looks strange:
```python
    except:
        raise
        parser.error("Git repository seems to have local modifications.")
```

scripts/import_srpm.py
```diff
@@ -64,110 +75,120 @@ def main():
...
+        else:
+            call_process(['git', 'checkout', args.branch])
+        call_process(['git', 'add', '--all'])
+
+        print(" committing...")
+        has_changes = False
+        try:
+            call_process(['git', 'diff-index', '--quiet', 'HEAD', '--'])
+        except Exception:
+            has_changes = True
```
ydirson, line 164: This too is a separate fix

ydirson (changes requested):

scripts/import_srpm.py
glehmann: So because github is bad at showing this kind of diff, I'm supposed to think about how the diff will be displayed in github and split my work in multiple commits? That seems a bit excessive to me.
Also note that the reviewer would still have to actually look at the diff which indents the code, to make sure nothing is changed.

glehmann:
There was even one more fix, which probably also need its own commit:

image

scripts/import_srpm.py
ydirson: Note I'm not talking about GH (because we all know how bad they are :wink:), but about reviewing with standard git tools.

scripts/import_srpm.py
```diff
@@ -64,110 +75,121 @@ def main():
...
+    # handle URL source
+    source_rpm = args.source_rpm
+    is_remote_rpm = is_url(source_rpm)
+
+    with TemporaryDirectory() if is_remote_rpm else nullcontext() as temp_dir:
+        if is_remote_rpm:
+            # get the src.rpm locally, and continue with the actual file
+            local_filename = f'{temp_dir}/{os.path.basename(source_rpm)}'
+            urlretrieve(source_rpm, local_filename)
+            source_rpm = local_filename
```
ydirson, line 87: The non-remote case could possibly be simplified: there seems to be no need to use a context manager at all in that case, if all the actions inside the `with` block are for the other case:
```suggestion
    if is_remote_rpm:
        with TemporaryDirectory() as temp_dir:
            # get the src.rpm locally, and continue with the actual file
            local_filename = f'{temp_dir}/{os.path.basename(source_rpm)}'
            urlretrieve(source_rpm, local_filename)
            source_rpm = local_filename
```

ydirson (changes requested):
otherwise ok

scripts/import_srpm.py
glehmann: I don't think it would work: the TemporaryDirectory context manager would delete the directory when exiting the `with` block, but we still need what this dir contains later in the code

ydirson: mb, looks like I read that with too little context :(

ydirson (approved):

PR #742, opened 2025-08-06
import_srpm: check command rpm2cpio|cpio

tperard:
`rpm2cpio` might not exist on the system, and `cpio` return the error "premature end of archive" in that case.

This doesn't check if `rpm2cpio` returns an error but it's probably ok to lean on `cpio` returning an error.

Also, check early if `rpm2cpio` actually exist on the system.

### DIFF
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -28,6 +28,11 @@ def main():
                 }[args.verbose]
     logging.basicConfig(format='[%(levelname)s] %(message)s', level=loglevel)
 
+    try:
+        call_process(['sh', '-c', 'command -v rpm2cpio'])
+    except subprocess.CalledProcessError:
+        parser.error("Command `rpm2cpio` missing")
+
     # check that the source RPM file exists
     if not os.path.isfile(args.source_rpm):
         parser.error("File %s does not exist." % args.source_rpm)
@@ -76,7 +81,7 @@ def main():
     print(" extracting SRPM...")
 
     os.chdir('SOURCES')
-    os.system('rpm2cpio "%s" | cpio -idmv' % source_rpm_abs)
+    call_process(['sh', '-c', 'rpm2cpio "%s" | cpio -idmv' % source_rpm_abs])
     os.chdir('..')
     os.system('mv SOURCES/*.spec SPECS/')
 


scripts/import_srpm.py
```diff
@@ -76,7 +76,7 @@ def main():
     print(" extracting SRPM...")
 
     os.chdir('SOURCES')
-    os.system('rpm2cpio "%s" | cpio -idmv' % source_rpm_abs)
+    call_process(['sh', '-c', 'rpm2cpio "%s" | cpio -idmv' % source_rpm_abs])
```
stormi, line 84: Could you also add a check higher in the script, so that we don't even attempt to do anything is rpmcpio is found missing?

tperard: Sure.
Is it ok to do it in posix shell compatible way?
`call_process(['sh', '-c', 'command -v rpm2cpio'])`
Or should I introduce a new dependency and use `shutil.which` ? But I'm not sure this will use the same algorithm as `sh`.

stormi: No strong opinion. Just make sure you catch the error to display a user-friendly message, so people don't have to decipher a backtrace.

scripts/import_srpm.py
```diff
@@ -76,7 +81,7 @@ def main():
...
```
glehmann, line 84:
```suggestion
    call_process(['sh', '-o', pipefail', '-c', 'rpm2cpio "%s" | cpio -idmv' % source_rpm_abs])
```

otherwise it won't fail if rpm2cpio fails.
Or we can use #743 — I didn't see there was already a PR for that problem 😅 
It's in pure Python, but it's a much bigger change.

scripts/import_srpm.py
```diff
@@ -28,6 +28,11 @@ def main():
                 }[args.verbose]
     logging.basicConfig(format='[%(levelname)s] %(message)s', level=loglevel)
 
+    try:
+        call_process(['sh', '-c', 'command -v rpm2cpio'])
+    except subprocess.CalledProcessError:
```
glehmann, line 33:
```suggestion
    if shutil.which('rpm2cpio') is None:
```
just the same, but with just python 🙂

scripts/import_srpm.py
tperard: I'm sorry, `pipefail` is a bash option, it doesn't exist on posix shell. I've actually look before writing my commit message, saying that this ignores error from `rpm2cpio`.
We don't know what implementation is used for `sh`, it might be `dash` on Debian, or `bash` on Arch Linux.

glehmann: Sorry, I didn't read your comment carefully enough.

Too bad, I was liking your version better than mine, for its conciseness.

It's probably better to go with #743 then.

dinhngtu: If you're calling `sh` explicitly anyway you can just replace with `/bin/bash` right? Unless calling out bash is unwanted.

stormi:
@tperard It seems to me that there are unsolved comments left on this PR of yours.

tperard:
Replaced by #743

PR #743, opened 2025-08-06
Use pure python code in place of os.system

glehmann:
os.system calls the user's shell, and is not directly checking for errors. This should raise an error when cpio or rpm2cpio fail, or if they can't be found.

### DIFF
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -2,11 +2,41 @@
 import argparse
 import logging
 import os
+import shutil
 import subprocess
+from glob import glob
+
 
 def call_process(args):
     logging.debug("$ %s", args)
     subprocess.check_call(args)
+
+def pipe_commands(*commands: list[str]) -> bytes:
+    if not commands:
+        raise ValueError("The 'commands' list cannot be empty.")
+    if any(not cmd for cmd in commands):
+        raise ValueError("All commands in the list must be non-empty.")
+
+    processes: list[subprocess.Popen[bytes]] = []
+    next_process_stdin = None
+
+    for command in commands:
+        process = subprocess.Popen(
+            command,
+            stdin=next_process_stdin,
+            stdout=subprocess.PIPE,
+        )
+        processes.append(process)
+        next_process_stdin = process.stdout
+
+    final_stdout, _final_stderr = processes[-1].communicate()
+
+    for cmd, process in zip(commands, processes):
+        process.wait()
+        if process.returncode != 0:
+            raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
+
+    return final_stdout
 
 def main():
     parser = argparse.ArgumentParser(description='Imports the contents of a source RPM into a git repository')
@@ -27,6 +57,10 @@ def main():
                 2: logging.DEBUG,
                 }[args.verbose]
     logging.basicConfig(format='[%(levelname)s] %(message)s', level=loglevel)
+
+    for dep in ['cpio', 'rpm2cpio']:
+        if shutil.which(dep) is None:
+            parser.error(f"{dep} can't be found.")
 
     # check that the source RPM file exists
     if not os.path.isfile(args.source_rpm):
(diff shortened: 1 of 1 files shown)


scripts/import_srpm.py
```diff
@@ -2,12 +2,45 @@
...
+def pipe_commands(*commands: list[str]) -> bytes:
+    if not commands:
+        raise ValueError("The 'commands' list cannot be empty.")
+    if any(not cmd for cmd in commands):
+        raise ValueError("All commands in the list must be non-empty.")
+
+    processes: list[subprocess.Popen[bytes]] = []
+    last_process_stdout = None
+
+    for i, cmd_args in enumerate(commands):
```
psafont: I don't see any usages of the variable `i`, why is enumerate used here?

glehmann: I used it, then dropped it while writing the function, but forgot it there.

This is the kind of thing that makes me like static analysis :)

psafont (approved):

scripts/import_srpm.py
```diff
@@ -2,12 +2,45 @@
...
     subprocess.check_call(args)
 
+def pipe_commands(*commands: list[str]) -> bytes:
+    if not commands:
+        raise ValueError("The 'commands' list cannot be empty.")
+    if any(not cmd for cmd in commands):
+        raise ValueError("All commands in the list must be non-empty.")
+
+    processes: list[subprocess.Popen[bytes]] = []
+    last_process_stdout = None
```
ydirson: naming it `next_process_stdin` would make it more accurate

scripts/import_srpm.py
```diff
@@ -2,12 +2,45 @@
...
+    for command in commands:
+        process = subprocess.Popen(
+            command,
+            stdin=last_process_stdout,
+            stdout=subprocess.PIPE,
+        )
+        processes.append(process)
+        # to prevent deadlocks if the current process finishes before the previous one.
+        if last_process_stdout:
+            last_process_stdout.close()
```
ydirson: I'm not 100% sure of the nature of the deadlock, but avoiding keeping 2 readers on the pipe is definitely a good idea.  Maybe describe it like "the new process does the reading from this pipe, drop our ref to it, notably preventing deadlocks"?

scripts/import_srpm.py
```diff
@@ -2,12 +2,45 @@
...
+        last_process_stdout = process.stdout
+
+    final_stdout, _final_stderr = processes[-1].communicate()
+
+    for i, p in enumerate(processes):
+        p.wait()
+        if p.returncode != 0:
+            raise subprocess.CalledProcessError(returncode=p.returncode, cmd=commands[i])
```
ydirson: Maybe more clear with `for cmd, p in zip(commands, processes)`?

ydirson (changes requested):

scripts/import_srpm.py
glehmann: I suppose it depends on which side of the pipe you're looking at 🙂

ydirson: in one way yes, but I'm rather seeing it as "in one case the name is always valid, while in the other there is no previous process when launching the first one"

scripts/import_srpm.py
glehmann: I wrote that based on an example I found, but I can't produce any deadlock, and I find plenty of examples that don't use it. I'll remove it.

scripts/import_srpm.py
glehmann: That's why it's a `IO[bytes] | None`.
But I'm also ok with the other name, I'll use it 👍

scripts/import_srpm.py
glehmann: done

scripts/import_srpm.py
glehmann: done

scripts/import_srpm.py
ydirson: It should be kept, as explained (much better) [in the API doc](https://docs.python.org/3/library/subprocess.html#replacing-shell-pipeline)

stormi:
What's the status? I see pushes so I assume that @glehmann would like a re-review from @ydirson, but that's not obvious without actually re-requesting the review.

PR #745, opened 2025-08-11
Require using --pre-build or --test-build when building with a user target

glehmann:


scripts/koji/koji_build.py
```diff
@@ -232,7 +240,7 @@ def main():
             + (['--nowait'] if is_nowait else [])
         )
         print('  '.join(command), flush=True)
-        subprocess.check_call(command)
+        # subprocess.check_call(command)
```
last-genius: Why is this commented out?

glehmann: that's a mistake, thanks for noticing :+1:

glehmann:
@last-genius do you have the last version? This would be the behavior I expect with the comment out line you reported, but which is fixed now.

scripts/koji/koji_build.py
```diff
@@ -50,6 +50,14 @@ def check_commit_is_available_remotely(dirpath, hash, target, warn):
...
         if not subprocess.check_output(['git', 'branch', '-r', '--contains', hash]):
             raise Exception("The current commit is not available in the remote repository")
+        try:
+            if target is not None and re.match(r'v\d+\.\d+-u-.+', target):
+                raise Exception(f"Building with a user target requires using --pre-build or --test-build.\n")
+        except Exception as e:
+            if warn:
+                print(f"warning: {e}", flush=True)
+            else:
+                raise e
```
psafont: The whole point of blocking this operation is because it can be dangerous. I would still like to see the exception being raised when warn is enabled:
```suggestion
            raise e
```

glehmann: I agree, I can't think of any use case where we may want to bypass this check.

glehmann: done

last-genius (approved):

PR #746, opened 2025-08-18
Add a script to terminate idle pg connections

glehmann:
Koji keeps opening new connections to the database, and keeps them open
up to the point where no new connection can be open. Koji then starts
to fail, with a database outage error message.

This scripts terminates the connexions older that 10 minutes, and is
intended to run every minute on the koji server.

### DIFF
--- a/scripts/terminate_idle_connections.py
+++ b/scripts/terminate_idle_connections.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+
+import psycopg2
+
+IDLE_TIMEOUT_MINUTES = 10
+
+def main():
+    with psycopg2.connect(dbname="postgres") as conn, conn.cursor() as cur:
+        sql_query = f"""
+        SELECT pg_terminate_backend(pid)
+        FROM pg_stat_activity
+        WHERE
+            state = 'idle'
+            AND NOW() - state_change > INTERVAL '{IDLE_TIMEOUT_MINUTES} minutes'
+            AND pid <> pg_backend_pid();
+        """
+        cur.execute(sql_query)
+        num_terminated = len(cur.fetchall())
+    if num_terminated > 0:
+        print(f"terminated {num_terminated} connections")
+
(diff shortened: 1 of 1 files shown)


nathanael-h (approved):

ydirson (changes requested):
There should be some context, what pg connections are we talking about?

ydirson (approved):
lgtm (not 100% sure though, whether that the `print()` before `commit()` should not be saying "terminating", but that should not have much impact :wink:)

glehmann:
Yes, that was a weird place to add it.
I've also used the connection as a context manager.

scripts/terminate_idle_connections.py
stormi, line 1: It looks like a temporary workaround, so I'm not entirely at ease with merging this and risking to make it a long lived hack.

Unless that's what the koji developers recommend?

nathanael-h: It is not what koji developers recommend at all. [Quoting](https://matrix.to/#/!taaLTunupmsbVpmVvP:fedora.im/$pkBewLJ4OfDN7j8mwiDQAGKJZFvoRMiY-E4eTAKGP9o?via=fedora.im&via=matrix.org&via=isidorus.fr) Mike McLean from : 


> Koji works on a one connection per thead model. If httpd leaves threads active, then they will keep their db connection.
> See discussion in https://pagure.io/koji/issue/4175
> 
> Issue #4175: Add an limit for the database connections - koji - Pagure.io
> 
> Most likely the solution to your issue lies in httpd configuration
> That said, it is normal for kojihub to keep a lot of connections open.

glehmann: It is indeed a temporary workaround.

stormi:
I'm converting this to draft. We probably don't want to spend more time on it unless we can't find a proper fix (and @nathanael-h might already have found one).

glehmann:
We don't need it anymore :tada:

PR #747, opened 2025-08-20
Assign some maintainers to some packages with frequent updates

glehmann:
in order to avoid pinging everybody on mattermost for each update

### DIFF
--- a/scripts/rpm_owners/packages.json
+++ b/scripts/rpm_owners/packages.json
@@ -4510,7 +4510,7 @@
     },
     "v6d": {
         "added_by": "XS",
-        "maintainer": "?",
+        "maintainer": "XAPI & Network",
         "note": "proprietary, replaced by xcp-featured",
         "package": "v6d",
         "status": "ignored"
@@ -4637,7 +4637,7 @@
     },
     "xapi-clusterd": {
         "added_by": "XS",
-        "maintainer": "?",
+        "maintainer": "XAPI & Network",
         "note": "proprietary (GFS2)",
         "package": "xapi-clusterd",
         "status": "ignored"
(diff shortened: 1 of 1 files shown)


scripts/rpm_owners/packages.json
```diff
@@ -4637,7 +4637,7 @@
     },
     "xapi-clusterd": {
         "added_by": "XS",
-        "maintainer": "?",
+        "maintainer": "XAPI & Network",
         "note": "proprietary (GFS2)",
         "package": "xapi-clusterd",
         "status": "ignored"
```
ydirson, line 4643: I understand for v6d and xapi-storage-plugins, for which we may want to check changes to see if our replacements need additional work (which we likely want reflected at some point in a more accurate status), but can't this one purely be ignored?

glehmann: Yes probably. We should refine the statuses to clarify what is still tracked, even if not packaged, and what is actually ignored.

scripts/rpm_owners/packages.json
```diff
@@ -4510,7 +4510,7 @@
     },
     "v6d": {
         "added_by": "XS",
-        "maintainer": "?",
+        "maintainer": "OS Platform & Release",
```
stormi: That's for XAPI team.

ydirson (approved):

stormi:
Reminder that a force-push to a PR doesn't stand for "please re-review now, I've taken your comments into account". An arbitrary number of pushes and commits can be made until a PR is ready for re-review, so we have to ask for re-review explicitly, or respond to comments saying "Done".

stormi (approved):

PR #748, opened 2025-08-21
Cache koji builds

glehmann:
in order to decrease the load on koji. With what's currently in the pipe, we decrease the number of koji calls from 65 to 37 with the cache.

### DIFF
--- a/scripts/pkg_in_pipe/pkg_in_pipe.py
+++ b/scripts/pkg_in_pipe/pkg_in_pipe.py
@@ -206,9 +206,9 @@ def tag_priority(tag):
     tag = tag.split('-')[-1]
     return TAG_ORDER.index(tag)
 
-def find_previous_build_commit(session, build_tag, build):
+def find_previous_build_commit(build_tag, build):
     """Find the previous build in an higher priority koji tag and return its commit."""
-    tagged = session.listTagged(build_tag, package=build['package_name'], inherit=True)
+    tagged = KOJI.listTagged(build_tag, package=build['package_name'], inherit=True)
     tagged = sorted(tagged, key=lambda t: (tag_priority(t['tag_name']), -t['build_id']))
     build_tag_priority = tag_priority(build_tag)
     tagged = [
@@ -216,7 +216,7 @@ def find_previous_build_commit(session, build_tag, build):
     ]
     if not tagged:
         return None
-    previous_build = session.getBuild(tagged[0]['build_id'])
+    previous_build = KOJI.getBuild(tagged[0]['build_id'])
     if not previous_build.get('source'):
         return None
     return parse_source(previous_build['source'])[1]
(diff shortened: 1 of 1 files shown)


glehmann:
It would be great if it could be validated quickly—this might have an impact on koji, but only once merged :slightly_smiling_face:

scripts/pkg_in_pipe/pkg_in_pipe.py
```diff
@@ -253,6 +253,14 @@ def find_pull_requests(gh, repo, start_sha, end_sha):
             prs.update(commit_prs)
     return sorted(prs, key=lambda p: p.number, reverse=True)
 
+def get_koji_build(session, build_id) -> dict:
+    cache_key = f'koji-build-{build_id}'
+    if not args.re_cache and cache_key in CACHE:
+        return cast(dict, CACHE[cache_key])
```
ydirson, line 269: the cast looks necessary just because `CACHE` itself cannot be properly type-hinted?

scripts/pkg_in_pipe/pkg_in_pipe.py
```diff
@@ -253,6 +253,14 @@ def find_pull_requests(gh, repo, start_sha, end_sha):
...
+    else:
+        build = session.getBuild(build_id)
+        CACHE.set(cache_key, build, expire=RETENTION_TIME)
+        return build
```
ydirson, line 273: instead of managing such a cache manually, wouldn't `get_koji_build = functools.cache(session.getBuild)` do the job?  I guess other uses of `CACHE` could also go that same way.

ydirson (changes requested):

nathanael-h (approved):

scripts/pkg_in_pipe/pkg_in_pipe.py
glehmann: yes :+1:

scripts/pkg_in_pipe/pkg_in_pipe.py
glehmann: that's a persistent cache, using `diskcache`. I don't think `functool` has anything similar.

scripts/pkg_in_pipe/pkg_in_pipe.py
ydirson: thinking twice about it, `CACHE` could be typed using a `TypeDict`?

scripts/pkg_in_pipe/pkg_in_pipe.py
ydirson: Ah, right.  Looks like this is a common pattern, though, and there are some libs for this, like:
- https://joblib.readthedocs.io/en/stable/
- https://github.com/python-cachier/cachier
- https://pypi.org/project/persist-cache/

scripts/pkg_in_pipe/pkg_in_pipe.py
glehmann: I don't think it would fit well.

I could `assert isinstance(…)` instead of casting if you prefer

scripts/pkg_in_pipe/pkg_in_pipe.py
glehmann: there is a `memoize()` decorator in `diskcache`, but it doesn't fit well with the command line option to refresh the cache

glehmann: or maybe I can set the `expire` property after creating the `CACHE` object. I'll check

stormi:
What's the status? Does this need a re-review by @ydirson, or changes prior to this?

scripts/pkg_in_pipe/pkg_in_pipe.py
glehmann: Apparently, I can't do that. I would have loved to use `CACHE.memoize()`, but it really doesn't fit well with the cache customization from the CLI.

stormi (approved):

stormi:
Given the scope of the PR and it's age, as well as the (absence) of risk - any issue caused by the PR would have low consequences and be fixed by Gaëtan promptly - I take it on me to merge.

PR #749, opened 2025-08-21
import_srpm: make most cli parameters optional

glehmann:
the repository path defaults to .
the local branch is used if not provided on the command line

### DIFF
--- a/scripts/import_srpm.py
+++ b/scripts/import_srpm.py
@@ -43,9 +43,9 @@ def pipe_commands(*commands: list[str]) -> bytes:
 def main():
     parser = argparse.ArgumentParser(description='Imports the contents of a source RPM into a git repository')
     parser.add_argument('source_rpm', help='local path to source RPM')
-    parser.add_argument('repository', help='local path to the repository')
-    parser.add_argument('parent_branch', help='git parent branch from which to branch')
-    parser.add_argument('branch', help='destination branch')
+    parser.add_argument('repository', default='./', nargs='?', help='local path to the repository')
+    parser.add_argument('parent_branch', nargs='?', help='git parent branch from which to branch')
+    parser.add_argument('branch', nargs='?', help='destination branch')
     parser.add_argument('tag', nargs='?', help='tag')
     parser.add_argument('-v', '--verbose', action='count', default=0)
     parser.add_argument('-p', '--push', action='store_true', help='pull and push')
@@ -92,8 +92,14 @@ def main():
 
     if args.push:
         call_process(['git', 'fetch'])
-    call_process(['git', 'checkout', args.parent_branch])
-    if args.push:
+    if args.parent_branch:
+        call_process(['git', 'checkout', args.parent_branch])
+    if (
+        args.push
+        # only pull if the branch is already tracked
+        and subprocess.call(['git', 'rev-parse', '--abbrev-ref', '--symbolic-full-name', '@{u}'],
+                            stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL) == 0
+    ):
         call_process(['git', 'pull'])
 
     print(" removing everything from SOURCES and SPECS...")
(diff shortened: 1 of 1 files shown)


scripts/import_srpm.py
```diff
@@ -11,9 +11,9 @@ def call_process(args):
 def main():
     parser = argparse.ArgumentParser(description='Imports the contents of a source RPM into a git repository')
     parser.add_argument('source_rpm', help='local path to source RPM')
-    parser.add_argument('repository', help='local path to the repository')
-    parser.add_argument('parent_branch', help='git parent branch from which to branch')
-    parser.add_argument('branch', help='destination branch')
+    parser.add_argument('repository', default='./', nargs='?', help='local path to the repository')
+    parser.add_argument('parent_branch', nargs='?', help='git parent branch from which to branch')
+    parser.add_argument('branch', nargs='?', help='destination branch')
```
ydirson, line 48: Maybe we could take this opportunity to use options for those?  IIUC, specifying `branch` cannot be done without passing the first 2 ones (and `parent_branch` is indeed needed much less often than `branch`)

ydirson, line 48: The default behavior when `branch` is not passed is not obvious.  Is it a good idea to have that one optional?

scripts/import_srpm.py
```diff
@@ -116,20 +123,21 @@ def main():
...
+        call_process(['git', 'push', '--set-upstream', 'origin', branch])
         if args.tag is not None:
             call_process(['git', 'push', 'origin', args.tag])
 
-    print(" switching to master before leaving...")
-
-    call_process(['git', 'checkout', 'master'])
+    if args.branch:
+        print(" switching to master before leaving...")
+        call_process(['git', 'checkout', 'master'])
```
ydirson, line 172: That "switching to master before leaving" is indeed a definitely bad idea in itself.  It could make sense to restore what the current branch is in the end, OTOH.
But then, there is that "merging to master" below, which then does require switching to master (which is not necessarily "switching back").

ydirson (changes requested):

scripts/import_srpm.py
glehmann: It uses the branch currently checked out in the repository.

glehmann: I'm not using those parameters, so I can't really say.
@stormi ?

scripts/import_srpm.py
glehmann: I would prefer that too. But so far, I've done the branch management by hand, so I can't really say.

scripts/import_srpm.py
ydirson: I see that, though:
- I kind of dislike the idea of making unchecked assumptions about the checked out branch ... very much especially if we advise people to `--push`, and the previous behaviour, or even the new one when using `--push`, is to leave `master` being the current branch when done
- if we still want it (and fix that inconsistency), it should appear in the param doc

Further more, since the idea is to get it run automatically soon, we don't gain much by making it optional.

stormi: I don't think it's a good idea to make the branch optional. And I don't think it's really useful to make the current path the default.

stormi: I get it that the CLI interface for this command is not optimal, but that's really one command for which we don't want people to spare some thinking before executing it. Hence the positional mandatory arguments. 

One argument that is confusing is the "parent branch" argument, which could be removed if we updated the docs so that people created the branch by hand when it didn't exist yet. "parent branch" is an argument that could be a candidate for being optional. But then I think it would be more confusing than leaving it mandatory. If you're confused by the parameter, then you don't really know what the script does, so better not use it.

The merge to master option should go. I mostly used it back when I didn't go through a PR for such merges, and it is now useless.

Coming back to the initial branch after doing the work would be nicer than going to master.

glehmann: Well, when running `git push`, `git` makes an unchecked assumption that this is currently the right branch. It doesn't force us to provide the branch name, even though we could push something wrong. It looks the same to me.

But yes, let's run it automatically, and we won't have to provide anything on the command line :slightly_smiling_face:

stormi:
@glehmann Did you mean to re-request review from Yann after your last push?

glehmann:
Abandoning it in favor of running it automatically, eventually

PR #752, opened 2025-08-28
owners: add vmfs6-tools

ydirson:


scripts/rpm_owners/packages.json
```diff
@@ -4579,6 +4579,13 @@
         "package": "vim",
         "status": "packaged"
     },
+    "vmfs6-tools": {
+        "added_by": "XCP-ng",
+        "maintainer": "OS Platform & Release",
+        "note": "as-is, not supported, not in theproduc, only in xcp-ng-labs",
```
stormi: Assuming that it's a typo and not a word I don't know :)

```suggestion
        "note": "as-is, not supported, not in the product, only in xcp-ng-labs",
```

PR #753, opened 2025-08-29
Add tk and bluez to packages.json, and update tix

stormi:
We had to import these to rebuild python2 and python3.

### DIFF
--- a/scripts/rpm_owners/packages.json
+++ b/scripts/rpm_owners/packages.json
@@ -277,6 +277,13 @@
         "package": "blktap",
         "status": "packaged"
     },
+    "bluez": {
+        "added_by": "XS",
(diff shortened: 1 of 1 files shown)


scripts/rpm_owners/packages.json
```diff
@@ -277,6 +277,13 @@
         "package": "blktap",
         "status": "packaged"
     },
+    "bluez": {
+        "added_by": "XS",
+        "maintainer": "OS Platform & Release",,
```
ydirson: the json parser will likely not like all those double commas

stormi: Oops. Fixed.

ydirson (approved):

PR #757, opened 2025-09-05
koji_build.py: wait for the internal repo to be up to date

stormi:
Add `--wait-repo` to `koji build` so that it starts the task with a waitrepo subtask which ensures that the internal repository for the build tag is up to date.

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -229,7 +229,7 @@ def main():
             check_commit_is_available_remotely(git_repos[0], hash, None if is_scratch else target, args.force)
         url = koji_url(remote, hash)
         command = (
-            ['koji', 'build']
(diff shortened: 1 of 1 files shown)


glehmann:
Is there a reason why it's not the default in koji?

stormi:
> Is there a reason why it's not the default in koji?

I'll tell you if they answer my questions. I found it documented in https://docs.pagure.org/koji/repo_generation/ which covers the new behaviour of kojira regarding repo generation, but it doesn't tell why it isn't the default.

Also, I've seen waitrepo tasks in some cases before some builds, shortly after the upgrade, so I'm not sure. But adding it seems to be an improvement to make sure we avoid bad surprises when doing a build shortly after another one.

glehmann (approved):

PR #758, opened 2025-09-10
Only clean old branches when using --pre-build or --test-build

glehmann:
This is to prevent koji_build to write to the repository while it wouldn't need to for a standard build.


ydirson (approved):
lgtm

PR #762, opened 2025-09-22
Add a few maintainers to packages without

stormi:
Mostly proprietary packages that we don't include, but it's good to have an associated team to assign plane cards to see if updates to such proprietary components may have an impact on other components.

Team leaders as reviewers for this PR.


stormi:
No team leader complained: merging.

PR #764, opened 2025-09-24
koji_build.py: fix XCP-ng release detection for test builds

stormi:
The current code assumes we're in a vX.Y-u- tag, but we can also do test builds in v8.3-lab.


stormi:
Tested by doing a --test-build towards `v8.3-lab` and towards `v8.3-u-stormi1`.

PR #765, opened 2025-10-08
Remove duplicated scripts

nathanael-h:
Eliminate scripts that were committed in two repositories to maintain a single source of truth.


stormi:
* The character limit for a git commit title is 70 chars. If you need more, you probably need to fill the body of the commit message. See https://docs.xcp-ng.org/project/development-process/commit-message-conventions/
* Where's the new location?

nathanael-h:
1. Commit msg fixed.
2. The Ansible internal repo used to track the deployment of Koji

stormi:
So we're moving public scripts to an internal repository? I'm not sure this is right.

nathanael-h:
I am not sure neither. The scripts are really specific for our infra needs, if someone else wants to build XCP-ng from sources I doubt the scripts would help in anything. Maybe it was not needed to have them in a public repo initially, but I have no strong opinion. I just want to fix the split brain.

stormi:
Our policy is: public by default, unless it has to be private. Maybe it's OK to move these scripts to an internal repository, but I would have preferred not to discover it with a PR that removes them from what I still considered their main location. I wasn't even aware that they had been duplicated.

glehmann:
A PR is a proposition of change, that can be rejected or accepted.
Your opinion has been explicitly requested to give you a chance to agree or disagree before committing to anything.

What would be your preferred way to propose a change?

stormi:
> A PR is a proposition of change, that can be rejected or accepted.

I already expressed myself about this. A draft is a proposition. A PR is saying "this is the right solution to me, please merge unless you have strong concerns about it". For the reviewer, that's not the same thing.

Here you're asking me to put my stamp on an already made move. That's not what I call a proposition. In some contexts, it's good enough. Here I had a need for a bit more context, and maybe even a quick talk about the move, that's all :)

stormi (approved):
I now have all the information I need about the internal repo, how the files will be managed on it, and since the move is already done and I have no strong enough reason to block it, approving.

PR #774, opened 2025-11-21
koji: Add supported git branches (eg: [8.3, master] for v8.3-* tags)

rzr:
make `koji_build` accept builds for v8.3 and from both `master` and `8.3`

Preparation for switch `master` to be "next" aka v9

To my understanding, the motivation of change is part of a bigger plan, to prepare a smooth transition to enable bleeding edge development on master branch. Of course maintenance will remain in v8.x branches.

Once the master branch needs to be stabilized then a new 9.0 branch can be open (this mean declared in PROTECTED_TARGETS, and also move the `master` branch from `8.3` target to `9.0` for a future cycle)

Meanwhile the development on master is still open and might become the base for the next cycle (TBD 9.1 or 10).

Note that `8.3` git branches are not created yet in repos, it would make sense to open them soon or later, ideally when master branches need to receive feature from next release. Coincidentaly thoses 8.3 branches will be used for maintenance of the LTS release and then developers/RE might consider back-porting fixes from master to 8.3.

What is unclear to me in the plan is when the new koji target/tag v9 will be defined, or if an `unstable` target will be added before, probably this will have to be coordinated with CI.

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -16,14 +16,16 @@ except ImportError:
 
 TIME_FORMAT = '%Y-%m-%d-%H-%M-%S'
 
-# target -> required branch
+# target -> allowed branch(es)
+# To lock a target, just assign an empty set: "{}"
+# Unlisted targets are not protected then any branch will be allowed
 PROTECTED_TARGETS = {
-    "v8.2-ci": "8.2",
-    "v8.2-fasttrack": "8.2",
-    "v8.2-incoming": "8.2",
-    "v8.3-ci": "master",
-    "v8.3-fasttrack": "master",
-    "v8.3-incoming": "master",
+    "v8.2-ci": {"8.2"},
+    "v8.2-fasttrack": {"8.2"},
+    "v8.2-incoming": {"8.2"},
+    "v8.3-ci": {"master", "8.3"},
+    "v8.3-fasttrack": {"master", "8.3"},
+    "v8.3-incoming": {"master", "8.3"},
 }
 
 @contextmanager
@@ -46,20 +48,30 @@ def check_git_repo(dirpath):
     with cd(dirpath):
         return subprocess.run(['git', 'diff-index', '--quiet', 'HEAD', '--']).returncode == 0
 
-def check_commit_is_available_remotely(dirpath, hash, target, warn):
+def check_commit_is_available_remotely(dirpath, sha, target, warn):
     with cd(dirpath):
-        if not subprocess.check_output(['git', 'branch', '-r', '--contains', hash]):
+        output = subprocess.check_output(['git', 'branch', '-r', '--contains', sha])
+        if not output:
             raise Exception("The current commit is not available in the remote repository")
+        logging.debug('Commit %s is contained in remote branch: %s', sha, output.decode().strip())
+        
         if target is not None and re.match(r'v\d+\.\d+-u-.+', target):
             raise Exception("Building with a user target requires using --pre-build or --test-build.\n")
         try:
-            expected_branch = PROTECTED_TARGETS.get(target)
-            if (
-                expected_branch is not None
-                and not is_remote_branch_commit(dirpath, hash, expected_branch)
-            ):
-                raise Exception(f"The current commit is not the last commit in the remote branch {expected_branch}.\n"
-                                f"This is required when using the protected target {target}.\n")
+            allowed_branches = PROTECTED_TARGETS.get(target)
+            if allowed_branches is None:
+                logging.debug('Target %s is not protected, any branch is allowed', target)
+            else:
+                for branch in allowed_branches:
+                    try:
+                        if is_remote_branch_commit(dirpath, sha, branch):
+                            logging.debug('Commit %s is on top of branch %s (target: %s)', sha, branch, target)
+                            break
+                    except Exception as e:
+                        logging.debug('Ignoring %s', e)
+                else:
+                    raise Exception(f"The current commit is not the last commit of any remote allowed branches: {allowed_branches}.\n"
+                                    f"This is required when using the protected target {target}.")
         except Exception as e:
             if warn:
                 print(f"warning: {e}", flush=True)
@@ -152,10 +164,20 @@ def push_bumped_release(git_repo, target, test_build_id, pre_build_id):
         return commit
 
 def is_remote_branch_commit(git_repo, sha, branch):
+    """ Args:
+           git_repo: URL of git repository
+           sha: Git hash
+           branch: remote branch in repository
+        Returns:
+           True if commit is on top of an assumed remote branch, false otherwise
+        Raises:
+           Exception: If branch is missing from repo, or not reachable
+    """
     with cd(git_repo):
-        remote_sha = (
-            subprocess.check_output(['git', 'ls-remote', 'origin', f'refs/heads/{branch}']).decode().strip().split()[0]
-        )
+        references = subprocess.check_output(['git', 'ls-remote', 'origin', f'refs/heads/{branch}']).decode().strip().split()
+        if not references:
+            raise Exception(f'Remote branch "{branch}" missing or unreachable')
+        remote_sha = references[0]
     return sha == remote_sha
 
 def build_id_of(name, candidate):


scripts/koji/koji_build.py
```diff
@@ -53,13 +53,17 @@ def check_commit_is_available_remotely(dirpath, hash, target, warn):
...
-            ):
-                raise Exception(f"The current commit is not the last commit in the remote branch {expected_branch}.\n"
-                                f"This is required when using the protected target {target}.\n")
+            available = False
+            supported_branches = PROTECTED_TARGETS.get(target)
+            if supported_branches:
+                for branch in supported_branches:
+                    available = is_remote_branch_commit(dirpath, hash, branch)
+                    if available:
+                        logging.info(f"commit {hash} is on top of branch {branch} (target: {target})")
```
glehmann: This should be at the `debug` level IMO

rzr: Updated, First I thought it would be informative to developer, then he can double check

scripts/koji/koji_build.py
```diff
@@ -153,9 +157,8 @@ def push_bumped_release(git_repo, target, test_build_id, pre_build_id):
 
 def is_remote_branch_commit(git_repo, sha, branch):
     with cd(git_repo):
-        remote_sha = (
-            subprocess.check_output(['git', 'ls-remote', 'origin', f'refs/heads/{branch}']).decode().strip().split()[0]
-        )
+        list = subprocess.check_output(['git', 'ls-remote', 'origin', f'refs/heads/{branch}']).decode().strip().split()
+        remote_sha = list[0] if list else None
```
ydirson: overloading builtin type `list` looks like a bad idea

scripts/koji/koji_build.py
```diff
@@ -53,13 +53,17 @@ def check_commit_is_available_remotely(dirpath, hash, target, warn):
...
-            expected_branch = PROTECTED_TARGETS.get(target)
-            if (
-                expected_branch is not None
-                and not is_remote_branch_commit(dirpath, hash, expected_branch)
-            ):
-                raise Exception(f"The current commit is not the last commit in the remote branch {expected_branch}.\n"
-                                f"This is required when using the protected target {target}.\n")
+            available = False
+            supported_branches = PROTECTED_TARGETS.get(target)
+            if supported_branches:
```
ydirson: `if supported_branches is not None` would leave the ability of forbidding any build on a branch by setting its branch list to `[]`, otherwise just just removing the last branch from the list would have a likely unexpected result.

scripts/koji/koji_build.py
```diff
@@ -53,13 +53,17 @@ def check_commit_is_available_remotely(dirpath, hash, target, warn):
...
+                for branch in supported_branches:
+                    available = is_remote_branch_commit(dirpath, hash, branch)
+                    if available:
+                        logging.debug(f"commit {hash} is on top of branch {branch} (target: {target})")
+                        break
+            if not available:
+                raise Exception(f"The current commit is not the last commit of any remote branches {supported_branches}.\n"
+                            f"This is required when using the protected target {target}.\n")
```
ydirson: rather than using `available` just for this check, `for: else:` could help.

(includes dropping final `\n` in exception string)

```suggestion
            supported_branches = PROTECTED_TARGETS.get(target)
            if supported_branches:
                for branch in supported_branches:
                    if is_remote_branch_commit(dirpath, hash, branch):
                        logging.debug(f"commit {hash} is on top of branch {branch} (target: {target})")
                        break
                else:
                    raise Exception(f"The current commit is not the last commit of any remote branches {supported_branches}.\n"
                                f"This is required when using the protected target {target}.")
            else:
                raise Exception(f"Protected target {target} has no allowed branches.")
```

scripts/koji/koji_build.py
ydirson: What about a separate commit explaining what this change does?
Looks like a robustness fix for non-existent branches, which even though necessary for this commit because the `8.3` branches are not yet created, would at least in theory have been useful independently ("in theory", because we actually don't intend to ever delete those 😉).

ydirson (changes requested):
In the commit message:
- you can drop the "To my understanding," prefix of that paragraph, your understanding is actually correct :)
- maybe the paragraph about 9.1 and 10 could be dropped, as well as the last "what is unclear to me"?

scripts/koji/koji_build.py
rzr: Unsure I get your point here, are you suggesting to add "is not None"  ?

No problem to add it, it will work but I think is not necessary, or am I missing something ?

case1: if supported_branches: 

- [] : if [] will be false and then will not attempt to iterate, and just throw the exception
- None: same as above

case2:  if supported_branches is not None: 
- [] , will iterate on empty list and confirm it is not available
- None: same as First case without the not None extra

> otherwise just just removing the last branch from the list would have a likely unexpected result.

Why unexpected results ? because if list is empty ?

Do you mean when supported_branches are [], right ? 

According to: https://peps.python.org/pep-0008/

```
For sequences, (strings, lists, tuples), use the fact that empty sequences are false:

# Correct:
if not seq:
if seq:
```

ydirson: > No problem to add it, it will work but I think is not necessary, or am I missing something ?

With `if supported_branches:` both `None` and `[]` are handled the same, that is an explicit empty list of branches will be synonymous to a target not being listed as protected.  The latter case denotes a non-protected target, so this `if` causes the check to be (correctly) skipped.  However, the semantics of having a target associated with a list of branch is that when we add/remove to the list it respectively causes more/less branches to be allowed.  And if `[]` is synonymous to `None`, removing the last branch, instead of removing that branch from the allowed ones, causes all branches to become valid (the "unexpected result" I was referring to).

scripts/koji/koji_build.py
rzr: I added an exception in other commit, i guess it is clearer now... 

Btw please check my commit message about the tuple in  remote_sha = "("

scripts/koji/koji_build.py
```diff
@@ -54,19 +54,18 @@ def check_commit_is_available_remotely(dirpath, hash, target, warn):
             raise Exception("Building with a user target requires using --pre-build or --test-build.\n")
         try:
             supported_branches = PROTECTED_TARGETS.get(target)
```
ydirson: > from user perspective both will fail

It should not: when the branch is not declared the check is skipped and the build should always be accepted (regardless of the branch), whereas when the set is empty it means no branch is allowed and the check should fail

ydirson (commented):
I'm not sure it is useful to keep separate commits here

scripts/koji/koji_build.py
rzr: ok then it is a new feature  we are introducing here, i can do it now

rzr:
> I'm not sure it is useful to keep separate commits here

sure I can squash them all , btw i have other commits that pylint the source, can they be part of this PR ? or later one

scripts/koji/koji_build.py
ydirson: yes, but that looks part of the new behaviour, the case could not happen when we just had a `dict[str, str]` - so like the rest I think it will be more clear not to have a separate commit for this

ydirson:
> > I'm not sure it is useful to keep separate commits here
> 
> sure I can squash them all , btw i have other commits that pylint the source, can they be part of this PR ? or later one

Changing what pylint does or dealing with pylint reports is a different topic, let's keep them separate so we don't hold this PR with remarks targetting the other topic

scripts/koji/koji_build.py
ydirson: I answered to a note on that somewhere already, it is a parenthesized expression not a tuple (would need a comma to make it a tuple).

rzr: true, i was missleaded by the [0] and thought it was some kind of hack. well nevermind it is gone now, let me just update the commit message of my revised branch.

rzr:
Thank you for your time involved in reviews.

Should I merge it now to push more linter changes I have for next review ?

stormi:
Yes, you can merge.

PR #775, opened 2025-11-28
koji: Lint koji_build.py file a bit and relax linter

rzr:
This is a first pass fixing important and low hanging fruits. Pylint now passing (with relaxed header).

It will be appreciated if future changes can pass linter, and feel free to make it strictier.

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -1,8 +1,13 @@
 #!/usr/bin/env python3
+# pylint: disable=logging-fstring-interpolation, line-too-long, broad-exception-raised, broad-exception-caught, missing-function-docstring # noqa
+
+"""Helper script to build xcp-ng package using koji service"""
+
 import argparse
 import logging
 import os
 import re
+import sys
 import subprocess
 from contextlib import contextmanager
 from datetime import datetime, timedelta
@@ -12,7 +17,7 @@ try:
     from specfile import Specfile
 except ImportError:
     print("error: specfile module can't be imported. Please install it with 'pip install --user specfile'.")
-    exit(1)
+    sys.exit(1)
 
 TIME_FORMAT = '%Y-%m-%d-%H-%M-%S'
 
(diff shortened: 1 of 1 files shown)


stormi:
Have you talked with Yann and Gaëtan about it? There may be a need for coordination with the broader linting efforts.

scripts/koji/koji_build.py
```diff
@@ -105,10 +131,12 @@ def local_branch(branch):
         # prev_branch is empty when the head was detached
         subprocess.check_call(['git', 'checkout', prev_branch or commit])
 
+
```
stormi, line 134: Doubling newlines is not a rule we chose to enforce at the moment.

scripts/koji/koji_build.py
```diff
@@ -144,8 +174,8 @@ def find_next_release(package, spec, target, test_build_id, pre_build_id):
     build_nb = sorted(build_nbs)[-1] + 1 if build_nbs else 1
     if test_build_id:
         return f'{spec.release}.0.{test_build_id}.{build_nb}'
-    else:
-        return f'{spec.release}~{pre_build_id}.{build_nb}'
+    return f'{spec.release}~{pre_build_id}.{build_nb}'
```
stormi, line 177: We avoided an else, but do we need to **enforce** that?

scripts/koji/koji_build.py
```diff
@@ -28,33 +33,42 @@
...
-        raise Exception("Directory %s doesn't exist" % dirpath)
+        raise Exception(f"Directory {dirpath} doesn't exist")
     return dirpath
 
+
 def check_git_repo(dirpath):
     """check that the working copy is a working directory and is clean."""
     with cd(dirpath):
-        return subprocess.run(['git', 'diff-index', '--quiet', 'HEAD', '--']).returncode == 0
+        return subprocess.run(['git', 'diff-index', '--quiet', 'HEAD', '--'], check=False).returncode == 0
```
stormi, line 61: What rule is this a result of?

scripts/koji/koji_build.py
rzr: ok i can revert this, fyi flake8 complained about that

scripts/koji/koji_build.py
rzr: Sure we can skip it

scripts/koji/koji_build.py:145:4: R1705: Unnecessary "else" after "return", remove the "else" and de-indent the code inside it (no-else-return)

scripts/koji/koji_build.py
rzr: scripts/koji/koji_build.py:49:15: W1510: 'subprocess.run' used without explicitly defining the value for 'check'. (subprocess-run-check)

stormi: Ok. The check would be smarter and see that we're checking the return code, but I suppose that if it exists, it's because many people do mistakes here.

glehmann:
We haven't updated the project configuration in this repository, as we did in most other python projects, like xcp-ng-tests or xcp-ng-build-env.
I think it's worth updating it to match the configuration in the other repositories.

PR #781, opened 2026-02-03
git-import-srpm: script to create branches in source repos from sRPMs revision.

casasnovas:
On Debian derivatives, you'll need:
 sudo apt-get install rpm

Simply run it inside a SRPM git repository and give it a revision (can be
tag, branch, sha1, ...), e.g.:

```bash
CODE_REPO_PATH=/path/to/linux/repo SRPM_REPO_PATH=/path/kernel/srpm/repo git-import-srpm XS-8.2.1
```
To import _all_ releases as branches, just pass `--all`:
```bash
git-import-srpm --all
```
By default, the `SRPM_REPO_PATH` is the current working directory.  If no
`CODE_REPO_PATH` is given, a default path is inferred that is
`${SRPM_REPO_PATH}/../(linux|xen|qemu)` (depending on the product).

The script was used import all branches of qemu, xen and the Linux kernel
released in current and prior XCP versions, e.g.:
 - xen branch example: https://github.com/xcp-ng/xen/tree/xen/xcpng-4.17.5-23.1/base
 - qemu branch example: https://github.com/xcp-ng/qemu-dp/tree/qemu/xcpng-4.2.1-5.2.15.2/base
 - linux branch example: https://github.com/xcp-ng/linux/tree/linux/xcpng-4.19.19-8.0.32.1/base

The script gives idempotent runs, reconstructed git sha1 will stay the same
across runs for a given srpm revision.  The script will not touch any of your
repository worktrees or indexes, so it is safe to run any time even if you are
working in the source repo (patches are applied separately in a temporary worktree).

Note that for each srpm revision, the script will create two branches in the source
repository, in the form:

```
<product>/(xcpng|xenserver)-<version>/(pre-base|base)
```

The `pre-base` ref will point to the upstream commit that was used as basis by
the source RPM before applying patches.  Effectively, the patch-set applied by
a source rpm can be rev-listed with `git log .../pre-base../base` which allows
see-ing how the patchset evolved over time.

Signed-off-by: Quentin Casasnovas <quentin.casasnovas@vates.tech>

### DIFF
--- a/scripts/git-import-srpm
+++ b/scripts/git-import-srpm
@@ -31,26 +31,43 @@
 # CODE_REPO_PATH is given, a default path is inferred that is
 # SRPM_REPO_PATH/../(linux|xen|qemu) (depending on the product).
 
+set -o pipefail
+
 THIS_SCRIPT="$(readlink -f "$0")"
 
 SRPM_REPO_PATH=${SRPM_REPO_PATH:-${PWD}}
-PRODUCT=$(basename ${SRPM_REPO_PATH}/SPECS/*.spec .spec)
+PRODUCT="$(basename "${SRPM_REPO_PATH}/SPECS/"*.spec .spec)"
 CODE_REPO_PATH="${CODE_REPO_PATH:-${SRPM_REPO_PATH}/../${PRODUCT/kernel/linux}/}"
 SPEC_FILE="SPECS/${PRODUCT}.spec"
-if [ "${PRODUCT}" = "kernel" -o "${PRODUCT}" = "qemu" ]; then
+if [[ ${PRODUCT} = "kernel" || ${PRODUCT} = "qemu" ]]; then
     RELEASE_PREFIX="v"
-elif [ "${PRODUCT}" = "xen" ]; then
+elif [[ ${PRODUCT} = "xen" ]]; then
     RELEASE_PREFIX="RELEASE-"
 else
     echo "Unknown package: '${PRODUCT}'"
     exit 1
 fi
 
+function check_spec_file() {
+    local worktree="$1"
+    local revision="$2"
+    local spec_file
+
+    spec_file="$(git -C "${worktree}" ls-tree --name-only "${revision}" SPECS/)"
+
+    if [[ ${spec_file} != "${SPEC_FILE}" ]]; then
+        echo "${SPEC_FILE} does not exist in ${worktree}:${revision}" 1>&2
+        exit 1
+    fi
+}
+
 function get_build_dir() {
     local worktree="$1"
     local revision="$2"
 
-    git -C ${worktree} show "${revision}:${SPEC_FILE}" | \
+    check_spec_file "${worktree}" "${revision}"
+
+    git -C "${worktree}" show "${revision}:${SPEC_FILE}" | \
         rpmspec --query --define "_topdir ${worktree}" --qf '%{name}-%{version}\n' /dev/stdin 2>> "${COMMAND_LOGS}" | \
         head -n 1
 }
@@ -59,40 +76,34 @@ function get_base_version() {
     local worktree="$1"
     local revision="$2"
 
-    if [ "$(git -C ${worktree} ls-tree --name-only ${revision} SPECS/)" != "${SPEC_FILE}" ]; then
-        return
-    fi
+    check_spec_file "${worktree}" "${revision}"
 
-    git -C ${worktree} show "${revision}:${SPEC_FILE}" | rpmspec -P /dev/stdin 2>> "${COMMAND_LOGS}" | awk "/^Version:/{print \"${RELEASE_PREFIX}\" \$NF}"
+    git -C "${worktree}" show "${revision}:${SPEC_FILE}" | rpmspec -P /dev/stdin 2>> "${COMMAND_LOGS}" | awk "/^Version:/{print \"${RELEASE_PREFIX}\" \$NF}"
 }
 
 function get_release() {
     local worktree="$1"
     local revision="$2"
 
-    if [ "$(git -C ${worktree} ls-tree --name-only ${revision} SPECS/)" != "${SPEC_FILE}" ]; then
-        return
-    fi
+    check_spec_file "${worktree}" "${revision}"
 
-    git -C ${worktree} show "${revision}:${SPEC_FILE}" | rpmspec -P /dev/stdin 2>> "${COMMAND_LOGS}" | awk '/^Release:/{print $NF}'
+    git -C "${worktree}" show "${revision}:${SPEC_FILE}" | rpmspec -P /dev/stdin 2>> "${COMMAND_LOGS}" | awk '/^Release:/{print $NF}'
 }
 
 function get_full_version() {
     local base_version="$1"
     local release="$2"
 
-    echo ${base_version#${RELEASE_PREFIX}}-${release}
+    echo "${base_version#"${RELEASE_PREFIX}"}-${release}"
 }
 
 function get_vendor() {
     local worktree="$1"
     local revision="$2"
 
-    if [ "$(git -C ${worktree} ls-tree --name-only ${revision} SPECS/)" != "${SPEC_FILE}" ]; then
-        return
-    fi
+    check_spec_file "${worktree}" "${revision}"
 
-    git -C ${worktree} show "${revision}:${SPEC_FILE}" | grep -q -e XCP-ng -e xcp \
+    git -C "${worktree}" show "${revision}:${SPEC_FILE}" | grep -q -e XCP-ng -e xcp \
         && echo "xcpng" \
         || echo "xenserver"
 }
@@ -101,15 +112,15 @@ function get_branch_name() {
     local worktree="$1"
     local revision="$2"
 
-    base_version=$(get_base_version "${worktree}" ${revision})
+    base_version=$(get_base_version "${worktree}" "${revision}")
 
-    release="$(get_release "${worktree}" ${revision})"
+    release="$(get_release "${worktree}" "${revision}")"
 
-    vendor=$(get_vendor "${worktree}" ${revision})
+    vendor=$(get_vendor "${worktree}" "${revision}")
 
     full_version="$(get_full_version "${base_version}" "${release}")"
 
-    if [ -z "${vendor}" -o -z "{full_version}" ]; then
+    if [[ -z ${vendor} || -z ${full_version} ]]; then
         echo "unknown_branch_name_for_rev_${revision}"
     else
         echo "${PRODUCT}/${vendor}-${full_version}/base"
@@ -126,8 +137,8 @@ function init_worktrees() {
 
     # Upstream stable base
     base_version=$(get_base_version "${srpm_repo_worktree}" HEAD)
-    if [ "${PRODUCT}" == "xen" ]; then
-        if [ "${base_version}" = RELEASE-4.13.0 ]; then
+    if [[ ${PRODUCT} = "xen" ]]; then
+        if [[ ${base_version} = "RELEASE-4.13.0" ]]; then
             base_version=85e1424de2dda
         fi
     fi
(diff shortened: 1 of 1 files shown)


scripts/git-import-srpm
```diff
@@ -0,0 +1,410 @@
+#!/bin/bash -eu
+#
+# tl;dr:
+# - To import a specific revision revision, provided you're in srpm repo:
```
fallen: `revision` is stated twice

fallen: in `the` srpm repo

scripts/git-import-srpm
```diff
@@ -0,0 +1,410 @@
...
+#
+#  CODE_REPO_PATH=/path/to/linux/repo SRPM_REPO_PATH=/path/kernel/srpm/repo git-import-srpm XS-8.2.1
+#
+# By default, the SRPM_REPO_PATH is the current working directory.  If no
+# CODE_REPO_PATH is given, a default path is inferred that is
+# SRPM_REPO_PATH/../(linux|xen|qemu) (depending on the product).
+
+THIS_SCRIPT="$(readlink -f "$0")"
+
+SRPM_REPO_PATH=${SRPM_REPO_PATH:-$(pwd)}
```
fallen: I'm curious why using `$(pwd)` instead of `$PWD` or `${PWD}` as the first will fork/exec IIUC

scripts/git-import-srpm
```diff
@@ -0,0 +1,410 @@
...
+    done | (SRPM_REPO_PATH="${SRPM_REPO_PATH}" CODE_REPO_PATH="${CODE_REPO_PATH}" xargs -n 1 -P $(nproc) ${THIS_SCRIPT})
+}
+
+COMMAND_LOGS=/tmp/git-import-srpm.${1}.debug.log
+
+test -f ${COMMAND_LOGS} && rm ${COMMAND_LOGS} || :
+
+if [ "$1" = "--all" ]; then
+    import_all_branches
+    exit $$
```
fallen: `$$`? or `$?` ?

scripts/git-import-srpm
```diff
@@ -0,0 +1,410 @@
...
+
+function import_all_branches() {
+    local -A revs_for_version
+
+    echo "[+] Gathering unique branches"
+
+    cd ${SRPM_REPO_PATH}
+
+    for revision in $(git rev-list --topo-order HEAD); do
+	local branch_name="$(get_branch_name $(pwd) ${revision} 2>> "${COMMAND_LOGS}")"
```
fallen: I usually like indenting the content of `for` loops
In that effect I usually put the `do` in its own line, aligned with the `for` and the remaining core of the loop indented.

scripts/git-import-srpm
```diff
@@ -0,0 +1,410 @@
...
+
+if [ "$1" = "--all" ]; then
+    import_all_branches
+    exit $$
+fi
+
+SRPM_REPO_REVISION=$1
+
+TMPDIR=$(mktemp -d -t src-import-xen-XXX)
+if [ -z "${DONT_DELETE_TMP_DIR:-}"]; then
```
fallen: is this a hidden smiley? :D

scripts/git-import-srpm
casasnovas: I'm fine either way, I supposed $(pwd) is a bit more portable at the cost of a fork for a built-in, which isn't really a valid point as I've been using a few bashism with variable substitutions in the script anyway :)  I'll change it.

scripts/git-import-srpm
casasnovas: Oh I was wondering why I got a bad return value earlier, thanks for spotting this :)

scripts/git-import-srpm
casasnovas: Weird that it doesn't show, in my editor it is definitely indented.  I'll converted everything to spaces to make sure we all see the code the same way.

fallen:
I don't have time for a full review today but I'll get back to this PR in the next days :)
It looks good anyway! Thanks!

scripts/git-import-srpm
fallen: something else, you need to add a space before the `]`

scripts/git-import-srpm
```diff
@@ -0,0 +1,385 @@
...
+
+    base_version=$(get_base_version "${worktree}" ${revision})
+
+    release="$(get_release "${worktree}" ${revision})"
+
+    vendor=$(get_vendor "${worktree}" ${revision})
+
+    full_version="$(get_full_version "${base_version}" "${release}")"
+
+    if [ -z "${vendor}" -o -z "{full_version}" ]; then
```
fallen: missing `$` for `full_version`

fallen:
I would generally encourage to run `shellcheck` on your script, it seems there are a few issues :)

casasnovas:
Oops that's embarassing, I thought my `lsp-mode` would run it in my editor, but I didn't have `shellcheck` installed so it silently didn't warn me about anything...

I should have fixed all `shellcheck` warnings/style/errors and info checks, and I've re-imported all `qemu` branches locally to make sure there was no breakage in functionality.

Let me know if there is anything else you'd like me to change, and thanks for the review!

If you're happy with it I'll squash the intermediate commits, I just did them to facilitate iterative reviews.

rzr:
This tool seems very similar to:

```
gbp import-srpm  --help
Usage: gbp import-srpm [options] /path/to/package.src.rpm [target]
```
https://honk.sigxcpu.org/piki/projects/git-buildpackage/

It works to some extends with RPM on my system, I can make a demo if curious.

casasnovas:
> This tool seems very similar to:
> 
> ```
> gbp import-srpm  --help
> Usage: gbp import-srpm [options] /path/to/package.src.rpm [target]
> ```
> 
> https://honk.sigxcpu.org/piki/projects/git-buildpackage/
> 
> It works to some extends with RPM on my system, I can make a demo if curious.

That looks nice, I didn't know it existed :)

Trying it on qemu isn't really successful here though:

```bash
quentin@laptop-tuxedo repos/qemu-rpms $ gbp import-srpm -v ./                                                                                                                                                                                                                                                                                         1
gbp:debug: Trying to import an unpacked srpm from './'
gbp:debug: Found spec file ./SPECS/qemu.spec
gbp:debug: ['git', 'rev-parse', '--show-cdup']
gbp:debug: ['git', 'rev-parse', '--is-bare-repository']
gbp:debug: ['git', 'rev-parse', '--git-dir']
gbp:debug: ['git', 'for-each-ref', '--format=%(refname:short)', 'refs/heads/']
gbp:debug: ['git', 'status', '--porcelain']
gbp:error: File '0001-usb-fix-setup_len-init-CVE-2020-14364.patch' listed in spec not found
```

The file is present inside SOURCES/ and I didn't find any options to make it look at the right place.

My main worry is that we'd still need to wrap it into something that does branch deduplication, branch name creation, git history stitching to upstream tags and likely it doesn't support idempotency (creating same git sha1 for same starting point + patch).  I'm happy to use it as the backend instead of the `apply_patch` function here if it does work though and we see any advantages.

Another tiny cons is the sheer amount of dependencies installing git-buildpackage brought in, here it's a tiny self-contained script.

rzr:
> > https://honk.sigxcpu.org/piki/projects/git-buildpackage/
> That looks nice, I didn't know it existed :)

It's coming from debian world and has been adapted to RPMs

I have been using it along gbs long time ago it was a great experience, see in action:

https://youtu.be/DLaD0hHY_F8?t=106#

 
> Trying it on qemu isn't really successful here though:

Let me give a try, I should be manageable.

> My main worry is that we'd still need to wrap it into something that does branch deduplication

unsure


> Another tiny cons is the sheer amount of dependencies installing git-buildpackage brought in, here it's a tiny self-contained script.

yes, It would requires extra distro integration (i am on it, in my idle time).

casasnovas:
@fallen any other comments?  I'm hoping to move forwards with this PR and potentially add automatic branch importing later.  If there are already existing tools that do the job better, I'm happy to use them in a future step.

rzr:
> If there are already existing tools that do the job better, I'm happy to use them in a future step.

Well better has to be defined but I managed to use gbp along qemu (need some teaks on my f43), feel free to call me if you want to reproduce, meanwhile here is the overall  flow:

# import

1st I run:

```
       gbp import-srpm \
            --create-missing-branches \
            --packaging-branch="${packaging_branch}" \
            --packaging-dir="${packaging_dir}" \
            --upstream-branch="${upstream_branch}" \
            --upstream-vcs-tag="$tag" \
            --vendor="$vendor" \
            --pristine-tar \
            --verbose \
            "${srpm_file}"
```

See history of:

- https://github.com/rzr/qemu/commits/packaging/master/

- Base is upstream tag
- Then a pristine tar patch (because we use tarball as sources) (this step depends on some packages not yet present on my f43)
- Then the packaging files

# Patch queue

From this we can generate a patch queue:

```
gbp pq-rpm --spec-file="${spec_file}"  import
```

See patches applied on upstream:

- https://github.com/rzr/qemu/commits/patch-queue/packaging/master


# Refresh patches (And spec)

And export the serie of patches back to packaging branch (with git context):
```
gbp pq-rpm --spec-file="${spec_file}" export --no-patch-numbers --verbose
```

- https://github.com/rzr/qemu/commits/packaging/devel/master

# Build again

Check, it will generate files again in SOURCES and SPECS


```
    gbp buildpackage-rpm \
        --packaging-dir="${packaging_dir}" \
        --git-packaging-branch="${packaging_branch}" \
        --git-upstream-branch="${upstream_branch}" \
        --git-spec-file="${spec_file}" \
        --git-native=no \
        --git-pristine-tar \
        --git-pristine-tar-commit \
        --git-force-create \
        --git-ignore-new \
        --git-no-build \
        --verbose \
```

Which is same inputs used in xcp-ng-rpm


In theory a srpm should be rebuildable from git too (WIP)

cc: @stormi @ydirson

fallen:
Thanks @rzr for sharing this other possible workflow.
@ydirson has created a meeting so that we can discuss further the topic of those two PRs.

casasnovas:
Thanks @rzr for going through with this!  From what I can tell, there is indeed a lack of idempotency when creating the branches, although maybe that could be work-arounded in the tool itself.

My take on this is that `git-import-srpm` is small, self-contained, without external dependencies, so easier to maintain overall, and provides a simpler interface while supporting idempotent runs and stitching of history in one single operation.  It also work in temporary worktrees so does not affect your local repos.  For all these reasons I'd go for KISS approach and keep `git-import-srpm`.

casasnovas:
Would it be acceptable if I moved this script into https://github.com/xcp-ng/hypervisor-dev, and just add some hooks to some github workflows to run it automatically for the packages we (Hypervisor&Kernel) support/maintain?

This way we (H&K team) can work on our own automation and explore ways to improve our developer experience without this being perceived as a shared process impacting any other teams.  And when the platform team works on this topic in the future, there won't be things in the way, and we can simply remove our hooks.

What do you guys think?

stormi:
> Would it be acceptable if I moved this script into https://github.com/xcp-ng/hypervisor-dev, and just add some hooks to some github workflows to run it automatically for the packages we (Hypervisor&Kernel) support/maintain?
> 
> This way we (H&K team) can work on our own automation and explore ways to improve our developer experience without this being perceived as a shared process impacting any other teams. And when the platform team works on this topic in the future, there won't be things in the way, and we can simply remove our hooks.
> 
> What do you guys think?

I think that's consistent with the outcome of the discussions we had on the topic all together.

ydirson:
Maybe the script should be renamed, today its name is quite close to existing `import_srpm.py` so the difference between them is not obvious to a fresh eye.

What about aligning with similar "import to git" scripts like `git-(arch|cvs|quilt)import` and go with `git-srpmimport`?  That does not change much, but at least the naming convention can help those familiar with it.

casasnovas:
PR moved to the hypervisor-dev repository:
https://github.com/xcp-ng/hypervisor-dev/pull/7

PR #782, opened 2026-02-05
git-review-rebase.py: add new script.

casasnovas:
This script is supposed to help review rebases by presenting the symmetric difference of rebased commits in two ranges.  Rebased commits are matched by multiple heuristics (in priority order: commit sha1, patch-ids, title) and it is possible to see a side-by-side diff of matched commits by pressing enter.  The list of commits can be filtered by match-type, or fuzzy-filtered based on commit sha1s or titles.

The side-by-side diff can be augmented with either full function context, or git blame info by pressing specific keys.

Comments on dropped commits can be added, so that the person doing the rebase can justify dropped commits.

Demo available at:

 https://asciinema.org/a/wDwcUTh6aJpvrvyX

Patch-ids are cached directly within the git repo, in specific refs, so that patch-ids caches can be shared amongst devellopes and don't have to be recomputed too often as it is a slow operation.

### DIFF
--- a/scripts/git-review-rebase/src/git_review_rebase/ui/diff_table.py
+++ b/scripts/git-review-rebase/src/git_review_rebase/ui/diff_table.py
@@ -172,7 +172,7 @@ class DiffPrettyRowMaker:
             return (Text(""), Text(""), Text(""))
 
         left_diff_line, middle_char, right_diff_line = (
-            line[: self.middle_point],
+            line[: self.middle_point].rstrip(),
             line[self.middle_point],
             line[self.middle_point + self.right_first_char_index :],
         )


scripts/git-review-rebase.py
```diff
@@ -0,0 +1,1223 @@
...
+import re
+import subprocess
+
+from collections import OrderedDict
+
+from difflib import SequenceMatcher
+
+from enum import Enum, StrEnum, Flag, auto
+
+from pathlib import Path
```
fallen: This seems unused

scripts/git-review-rebase.py
```diff
@@ -0,0 +1,1223 @@
...
+
+from pygit2.enums import SortMode
+from pygit2 import Walker
+
+from pygments.lexers import get_lexer_for_filename
+
+from rich.text import Text
+from rich.style import Style
```
fallen: This seems unused.

casasnovas:
> Nice PR! Thanks!
> 
> Does this need new dependencies that should be added to the `pyproject.toml` or somewhere else?

Actually, let me put this in its own repository, properly packaged (and with all python linters happy :).

I wanted to share this quickly as I referenced it in another PR and it was growing out of control in terms of features, but I'll take the time to clean this up.

fallen:
> > Nice PR! Thanks!
> > Does this need new dependencies that should be added to the `pyproject.toml` or somewhere else?
> 
> Actually, let me put this in its own repository, properly packaged (and with all python linters happy :).
> 
> I wanted to share this quickly as I referenced it in another PR and it was growing out of control in terms of features, but I'll take the time to clean this up.

Ok indeed maybe its own repo would be better so that it has its own `pyproject.toml` and everything.
Note that it could just be in a directory as well, like the koji tooling.

casasnovas:
> > Actually, let me put this in its own repository, properly packaged (and with all python linters happy :).
> > I wanted to share this quickly as I referenced it in another PR and it was growing out of control in terms of features, but I'll take the time to clean this up.
> 
> Ok indeed maybe its own repo would be better so that it has its own `pyproject.toml` and everything. Note that it could just be in a directory as well, like the koji tooling.

Alright I've split into multiple python modules and used pyproject.toml, so it should be ready for review.  It passes flake8, ruff check, black and mypy.

glehmann:
Just gave a quick look for now, but I'll review it soon!

We have https://github.com/xcp-ng/python-project-template/ to help keep our projects to similar standards. Could you use it here?

casasnovas:
> Just gave a quick look for now, but I'll review it soon!
> 
> We have https://github.com/xcp-ng/python-project-template/ to help keep our projects to similar standards. Could you use it here?

Done!  I adapted the workflows so that it should run only for the git-review-rebase project and will let other python scripts integrate as it's out-of-scope for this PR :)

casasnovas:
To test the script:
```bash
#  Install the script
cd xcp/scripts/git-review-rebase
pip install -e ./

# Clone a linux repo and use the script on it
git clone git@github.com/xcp-ng/linux.git
cd linux


# This will save you 5min on the first run of the script to calculate the patch-ids
git fetch origin refs/patchids/\*:refs/patchids/\* || cat patch-ids-packed-refs.txt >> .git/packed-refs

git-review-rebase origin/kernel/xcpng-4.19.19-8.0.44.1/pre-base..origin/kernel/xcpng-4.19.19-8.0.44.1/base origin/kernel/xcpng-4.19.325-8.0.44.1/pre-base..origin/kernel/xcpng-4.19.325-8.0.44.1/base
```

[patch-ids-packed-refs.txt](https://github.com/user-attachments/files/25228739/patch-ids-packed-refs.txt)

casasnovas:
Folks,

Given this is a completely new script and it does passes all of the linters and test usage, is it OK to move forward with the merging?  I am happy to add features/change things as separate PRs anyway.

Cheers,
Q

stormi:
> Folks,
> 
> Given this is a completely new script and it does passes all of the linters and test usage, is it OK to move forward with the merging? I am happy to add features/change things as separate PRs anyway.
> 
> Cheers, Q

You mean, without a review by a human?

casasnovas:
> > Folks,
> > Given this is a completely new script and it does passes all of the linters and test usage, is it OK to move forward with the merging? I am happy to add features/change things as separate PRs anyway.
> > Cheers, Q
> 
> You mean, without a review by a human?

I just mean I don't think it needs a deep/extended review, but more of a quick pass on the code and test usage given this is a completely new script, completely isolated from the rest of the repo and self-contained.

casasnovas:
Following discussions with Gaetan and the hypervisor team, who talked with the platform team, I am moving this script to its own repository - I mistakenly assumed this repository was hosting all dev tools and I think this is causing confusion.

stormi:
I don't object to a separate repository, but I don't understand why it we deemed required to move it outside a repository which historically hosted developer tools. I would have rather waited for the planned discussion on Monday, whose goal was specifically to discuss how to move on, before taking action.

casasnovas:
> I don't object to a separate repository, but I don't understand why it we deemed required to move it outside a repository which historically hosted developer tools. I would have rather waited for the planned discussion on Monday, whose goal was specifically to discuss how to move on, before taking action.

I can always re-open this PR if the outcome of the meeting is that it should live here, but I personally felt that the tool deserved its own repository because:
- it isn't (and won't be) integrated with our infrastructure
- it is standalone, self-contained and is useful for a variety of use-cases when doing rebases (outside of an XCP-centric) workflow
- Easier to track releases/versioning issues that are specific to the tool in its own repository (and I have a few enhancements in mind already).

My original thought was to have a dedicated repository for this but I thought it would be more trouble to ask for rights to create it and own it (i.e. I was being lazy and assuming extra process :).  I was happily surprised to see it's not the case and I could simply click a button to create one, so I just did :-)  Again, if the outcome is that it should live here, I can simply remove the new repository.

Now, on top of all of these reasons, I am feeling a bit of friction where I wasn't expecting any for a new script that doesn't have any effect on anything/anyone (we're more than three weeks into the PR, and we have a meeting to discuss about it, that does a feel a bit much).

I genuinely find it useful, and think others might as well (although nobody has to use it), so rendering it accessible ASAP sounds to me important - I see it as an enabler for the coming `v4.19` review that I do expect to take quite a bit of time given the sheer amount of changes.

stormi:
I was just questioning the idea of moving to a separate repo to avoid confusion without waiting for today's meeting, when I don't think there should be confusion related to where the script lives, which is a detail.

PR #784, opened 2026-02-12
Add Base for imported rpm

rzr:
The most recent import is listed, It may be useful for consequent imports, to align timeline and preserve history, context.

Currently information is stored in note field, and maybe relocated elsewhere (and schema updated).

Base can be changed if necessary, and updated accordingly, feel free also link related PRs, to keep trace of the motivation.

Only packages involved in openssl-3 transition are listed. in this change.

Relate-to: https://github.com/xcp-ng-rpms/libarchive/pull/1
Relate-to: https://github.com/xcp-ng-rpms/trousers/pull/1
Relate-to: https://github.com/xcp-ng-rpms/iputils/pull/1
Relate-to: https://github.com/xcp-ng-rpms/python-pycurl/pull/1
Relate-to: https://github.com/xcp-ng-rpms/ipmitool/pull/1
Relate-to: https://github.com/xcp-ng-rpms/ssmtp/pull/1

### DIFF
--- a/scripts/rpm_owners/packages.json
+++ b/scripts/rpm_owners/packages.json
@@ -250,6 +250,7 @@
     "bind": {
         "added_by": "XS",
         "maintainer": "OS Platform & Release",
+        "note": "Base: bind-9.9.4-63.xs8.src.rpm",
         "package": "bind",
         "status": "packaged"
     },
@@ -274,6 +275,7 @@
     "blktap": {
         "added_by": "XS",
         "maintainer": "Storage",
+        "note": "Base: blktap-3.55.5-6.xs8.src.rpm",
         "package": "blktap",
         "status": "packaged"
     },
@@ -456,7 +458,7 @@
     "cmake3": {
         "added_by": "XCP-ng",
         "maintainer": "OS Platform & Release",
-        "note": "Initially imported by us, XS started using it. For XS 8, they forked cmake-3.26.4-4.fc37.src.rpm. At first, we keep ours because we found nothing that would require cmake3 in recent updates",
+        "note": "Base: cmake3-3.26.4-3.xs8.src.rpm , Initially imported by us, XS started using it. For XS 8, they forked cmake-3.26.4-4.fc37.src.rpm. At first, we keep ours because we found nothing that would require cmake3 in recent updates",
         "package": "cmake3",
         "status": "packaged"
     },
@@ -514,6 +516,7 @@
     "coreutils": {
         "added_by": "XS",
         "maintainer": "OS Platform & Release",
+        "note": "Base: coreutils-8.22-22-xs8.src.rpm",
         "package": "coreutils",
         "status": "packaged"
     },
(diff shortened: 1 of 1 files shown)


scripts/rpm_owners/packages.json
```diff
@@ -4079,6 +4102,7 @@
     "rsync": {
         "added_by": "XS",
         "maintainer": "OS Platform & Release",
+        "note": "Base: 8.3-import_rsync-3.4.1-1.xs8.src.rpm",
```
fallen: Is it normal that the package name is prefixed by `8.3-import_`?

fallen:
Do I understand correctly that when rebasing our SRPMs on XS updates we should then update the `packages.json` file of this repo from now on?

rzr:
> Do I understand correctly that when rebasing our SRPMs on XS updates we should then update the `packages.json` file of this repo from now on?

Not everytime, only when the base is changing  (from upstream to xs or the other way arround)

fallen:
> > Do I understand correctly that when rebasing our SRPMs on XS updates we should then update the `packages.json` file of this repo from now on?
> 
> Not everytime, only when the base is changing (from upstream to xs or the other way arround)

Yes, but I mean, the package name will get outdated by new XS releases, right?

For instance:

```
"note": "Base: bind-9.9.4-63.xs8.src.rpm",
```

Here the bind package version information will get outdated, won't it?

rzr:
> Yes, but I mean, the package name will get outdated by new XS releases, right?

yes, but base will remain the same

rzr:
> What does having the info here bring us? Don't we have already the info with the `XS-*` and `upstream-*` branches, without the need for more manual steps when we update?

yes the info is in branches,  and it is not supposed to be updated each time, I was suggested to provide some hints for packagers to know where to look for, for upcoming updates and ideally minimize divergence to stream of the base (set once).

ydirson:
> I was suggested to provide some hints for packagers to know where to look for, for upcoming updates 

What info is not already included by looking at the merge commits and the upstream branches?

> and ideally minimize divergence to stream of the base (set once).

How does it help, over checking divergence from upstream by diffing with the upstream branch?

rzr:
> What info is not already included by looking at the merge commits and the upstream branches?

There is no additional information, except the motivation in commit message or valuable information that useful for  further works.
 
> How does it help, over checking divergence from upstream by diffing with the upstream branch?

This is a synthetic report of what base are we using, it can help for stats or tracking strategy changes

stormi:
> How does it help, over checking divergence from upstream by diffing with the upstream branch?

It helps greatly to have a note that says how we manage a given package, that is then automatically embedded in work items created whenever a new XS SRPM is released.

How the branch history looks like is not something everyone who may intervene in packaging (that is, not just the platform tem) will easily understand.

That's why I suggested to add a note here. 

However I wouldn't put the exact version, that will not age well, but rather a simple description:
* Based on upstream-8.3, watch diverging XS-8.3 package for important changes.
* Based on XS-8.3 itself based on upstream-8.3. Watch upstream for important patches.
* Based on XS-8.3's own packaging from sources.
* XCP-ng packaging, watch diverging XS-8.3 package for important changes.

Etc.

Else, we could have such metadata directly in the repositories, but that needs to be in an exploitable form (an enum value that we can automatically gather + free additional descriptive text?)

PR #788, opened 2026-03-10
koji_build: support for specfile-in-topdir layout

ydirson:
This is necessary for packages out of almalinux git repo

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -166,7 +166,7 @@ def push_bumped_release(git_repo, target, test_build_id, pre_build_id):
     t = datetime.now().strftime(TIME_FORMAT)
     branch = f'koji/test/{test_build_id or pre_build_id}/{t}'
     with cd(git_repo), local_branch(branch):
-        spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec']).decode().splitlines()
+        spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec', '*.spec']).decode().splitlines()
         assert len(spec_paths) == 1
         spec_path = spec_paths[0]
         with Specfile(spec_path) as spec:


scripts/koji/koji_build.py
```diff
@@ -164,6 +164,8 @@ def push_bumped_release(git_repo, target, test_build_id, pre_build_id):
     branch = f'koji/test/{test_build_id or pre_build_id}/{t}'
     with cd(git_repo), local_branch(branch):
         spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec']).decode().splitlines()
```
glehmann: I think you can do it with a single command:

```suggestion
        spec_paths = subprocess.check_output(['git', 'ls-files', 'SPECS/*.spec', '*.spec']).decode().splitlines()
```

rzr: which one should come first ?

BTW it is common to also have packaging/*.spec (in upstream packages), I would add it too

like: https://github.com/agx/git-buildpackage/tree/master/packaging

ydirson: > which one should come first ?

I don't think it matters much, there should never be both.  Gaetan's suggestion should take care of this if it ever happens.

> BTW it is common to also have packaging/*.spec (in upstream packages), I would add it too
> 
> like: https://github.com/agx/git-buildpackage/tree/master/packaging

Given koji is supposed to be phased out at some point, I'd rather not add things that we are likely to never need. We can add it anyway if it becomes necessary, and when we are able to test it.

ydirson: applied

glehmann (approved):

scripts/koji/koji_build.py
ydirson: Actually, none of the proposal are indeed correct, as this degenerate case shows:
```
user@work-vates:xen (ydi/9 =)$ git ls-files '*.spec'
SOURCES/xen.spec
SPECS/xen.spec
```

Actually I can't find find an invocation that would do what we want.  Can we do anything else than a posteriori filtering?

glehmann: Ho, that's unfortunate for our use case. I don't see any option to match only the root level in `git ls-file`.

stormi:
Ping @rzr.

stormi (changes requested):
Seen with Yann: needs to be done better following Gaëtan's remark.

PR #790, opened 2026-03-11
Add a script to copy a branch to another branch in all our rpm repositories

glehmann:
And also move the PR destination branch matching the old branch to the new branch

### DIFF
--- a/scripts/rpm_owners/copy_branch.py
+++ b/scripts/rpm_owners/copy_branch.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+
+import github as gh
+from github.Repository import Repository
+
+
+def copy_branch(repo: Repository, source: str, dest: str) -> bool:
+    source_branch = repo.get_branch(source)
+    try:
+        repo.create_git_ref(ref=f'refs/heads/{dest}', sha=source_branch.commit.sha)
+    except Exception as e:
+        print(f"Error creating branch {dest} in {repo.name}: {e}")
+        return False
+
+    return True
+
+def main():
+    parser = argparse.ArgumentParser(description="Copy a branch to another branch in all the active rpm repositories")
+    parser.add_argument("source", help="Source branch name")
+    parser.add_argument("dest", help="Destination branch name")
+    args = parser.parse_args()
+
+    # load the rpm data from the ref file
+    with open('packages.json') as f:
+        rpms = json.load(f)
+
+    auth = gh.Auth.Token(os.environ['GITHUB_TOKEN'])
+    g = gh.Github(auth=auth)
+    org = g.get_organization('xcp-ng-rpms')
+
+    gh_repos = {r.name: r for r in org.get_repos()}
+
+    pkgs = set(gh_repos.keys()) & set(rpms.keys())
+
+    ok = True
+    for pkg in pkgs:
+        repo = gh_repos[pkg]
+        ok &= copy_branch(repo, args.source, args.dest)
+    if not ok:
+        exit(1)
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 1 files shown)


stormi:
@glehmann If still useful let's ask the team for review.

olivierh-pro (approved):
LGTM! I would have wrapped it in a :

```python
if __name__ == "__main__":
    main()
```

scripts/rpm_owners/copy_branch.py
```diff
@@ -0,0 +1,53 @@
...
+with open('packages.json') as f:
+    rpms = json.load(f)
+
+auth = gh.Auth.Token(os.environ['GITHUB_TOKEN'])
+g = gh.Github(auth=auth)
+org = g.get_organization('xcp-ng-rpms')
+
+gh_repos = dict((r.name, r) for r in org.get_repos())
+
+pkgs = set(gh_repos.keys()).intersection(set(rpms.keys()))
```
vxgmichel, line 46:
```suggestion
pkgs = set(gh_repos) & set(rpms)
```

scripts/rpm_owners/copy_branch.py
```diff
@@ -0,0 +1,53 @@
...
+
+ok = True
+for pkg in pkgs:
+    repo = gh_repos[pkg]
+    ok &= copy_branch(repo, args.source, args.dest)
```
vxgmichel, line 51:
```suggestion
ok = all(copy_branch(gh_repos[pkg], args.source, args.dest) for pkg in pkgs))
```

scripts/rpm_owners/copy_branch.py
```diff
@@ -0,0 +1,53 @@
...
+
+# load the rpm data from the ref file
+with open('packages.json') as f:
+    rpms = json.load(f)
+
+auth = gh.Auth.Token(os.environ['GITHUB_TOKEN'])
+g = gh.Github(auth=auth)
+org = g.get_organization('xcp-ng-rpms')
+
+gh_repos = dict((r.name, r) for r in org.get_repos())
```
vxgmichel, line 44:
```suggestion
gh_repos = {r.name: r for r in org.get_repos()}
```

scripts/rpm_owners/copy_branch.py
```diff
@@ -0,0 +1,53 @@
...
+        repo.create_git_ref(ref=f'refs/heads/{dest}', sha=source_branch.commit.sha)
+    except Exception as e:
+        print(f"Error creating branch {dest} in {repo.name}: {e}")
+        return False
+
+    # Move PRs targeting the source branch to the new destination branch
+    try:
+        prs = repo.get_pulls(state='open', base=source)
+        for pr in prs:
+            if pr.base.ref == source:
```
vxgmichel, line 23: Is that check necessary since you already used `base=source` in `get_pulls`?

vxgmichel (approved):
LGTM 👍  Just a couple of stylistic suggestions, feel free to ignore if they don't feel like an improvement.

Also I agree with @olivierh-pro about the use of `if __name__ == "__main__":`, for the reason I formulated [here](https://github.com/xcp-ng/xcp/pull/823#issuecomment-4957413052):
> Some tools might try to import modules as part of their discovery process, and it's hard to know exactly which ones are going to run in this repository (considering that each developer might use different tools).

scripts/rpm_owners/copy_branch.py
glehmann: I don't think this is equivalent: all aborts early when it reaches a falsy value

scripts/rpm_owners/copy_branch.py
glehmann: I really don't know why I don't use dict comprehension, they look much better 👍

scripts/rpm_owners/copy_branch.py
glehmann: Indeed, it's redundant

scripts/rpm_owners/copy_branch.py
vxgmichel: Oh definitely, I missed that.

vxgmichel (approved):
All my comments have been addressed, re-approving for good measure 👍

PR #791, opened 2026-03-12
Fix non-incremented release when using test builds

glehmann:
The koji format seems to have changed and don't include the .xcpng8.3 at the end anymore.
The fix is still compabtible with the previous format, in case it depends on the client version, so it should continue to work for all the devs.

### DIFF
--- a/scripts/koji/koji_build.py
+++ b/scripts/koji/koji_build.py
@@ -148,12 +148,10 @@ def find_next_release(package, spec, target, test_build_id, pre_build_id):
         base_nvr = f'{package}-{spec.version}-{spec.release}.0.{test_build_id}.'
     else:
         base_nvr = f'{package}-{spec.version}-{spec.release}~{pre_build_id}.'
-    # use a regex to match %{macro} without actually expanding the macros
-    base_nvr_re = (
-        re.escape(re.sub('%{.+}', "@@@", base_nvr)).replace('@@@', '.*')
-        + r'(\d+)'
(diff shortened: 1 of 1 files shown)


scripts/koji/koji_build.py
```diff
@@ -151,8 +151,7 @@ def find_next_release(package, spec, target, test_build_id, pre_build_id):
     # use a regex to match %{macro} without actually expanding the macros
     base_nvr_re = (
         re.escape(re.sub('%{.+}', "@@@", base_nvr)).replace('@@@', '.*')
-        + r'(\d+)'
-        + re.escape(f'.xcpng{xcpng_version(target)}')
+        + r'(\d+)(' + re.escape(f'.xcpng{xcpng_version(target)}') + '| )'
```
ydirson: that expression starts to be quite complex to read, it would be useful to insert some comments to guide the reader

ydirson (changes requested):
It's not that the format has changed, but `xcp-ng-release` needs to be built for that logic to trigger, so there are 2 packages that do not obey this logic (release, and its build-dependency `branding-xcp-ng`).  Commit message should be adjusted.

scripts/koji/koji_build.py
glehmann: I've simplified the regex and updated the comment

stormi:
Ping @ydirson, review overdue on this one.

ydirson (approved):

stormi (approved):

PR #792, opened 2026-03-12
Add XCP-ng 9.0 in the report

glehmann:


scripts/pkg_in_pipe/pkg_in_pipe.py
```diff
@@ -197,7 +196,10 @@ def filter_issues(issues, urls):
 def tag_priority(tag):
     # drop the version in the tag — v8.3-incoming -> incoming
     tag = tag.split('-')[-1]
-    return TAG_ORDER.index(tag)
+    if tag in TAG_ORDER:
+        return TAG_ORDER.index(tag)
+    else:
+        return -1
```
ydirson, line 202: How is that related to `v9.0-incoming`?

BTW, looking at that code, I think `find_previous_build_commit` could avoid relying on hardcoded lists, and use `koji list-tag-inheritance` to get the actual relation to the tags it inherits from

glehmann: > How is that related to `v9.0-incoming`?

This function is used with the tag `almalinux-10.0`, which I think comes from the previous task that imported the packages.

> BTW, looking at that code, I think `find_previous_build_commit` could avoid relying on hardcoded lists, and use `koji list-tag-inheritance` to get the actual relation to the tags it inherits from

I don't think `find_previous_build_commit` is using a hardcoded list. Could you elaborate?

stormi: Ping @ydirson we are waiting for your answer here.

ydirson: > > BTW, looking at that code, I think `find_previous_build_commit` could avoid relying on hardcoded lists, and use `koji list-tag-inheritance` to get the actual relation to the tags it inherits from
> 
> I don't think `find_previous_build_commit` is using a hardcoded list. Could you elaborate?

I was referring to `TAG_ORDER` itself: it is trying to replicate the relationship existing as tag-inheritance.

But this is actually not a reason to block this PR.

ydirson (approved):

stormi (approved):

PR #795, opened 2026-03-20
ghcr-cleanup: expand-packages: true requires a PAT

glehmann:


stormi (approved):

stormi:
@glehmann I think I was letting you merge as this is a very simple commit and no one else was asked for review. Let's merge if still valid?

PR #796, opened 2026-03-23
Fix missing plane cards

glehmann:
The script was not using the pagination and missed the oldest cards


rzr (approved):
thank you, going to test this

btw: i would use constants as def parms :)

stormi (approved):

PR #803, opened 2026-04-14
github: Add json lint check

rzr:
## Motivation

This will prevent future breaks (like the previous one that @glehmann fixed),
more check to be introduced (on yaml files etc)

## Demo:

Failed check:

https://github.com/xcp-ng/xcp/actions/runs/24389050719/job/71229969217

```
info: check: will pass if sources are linted, if not please lint and commit
git diff --stat --exit-code
 scripts/rpm_owners/packages.json | 10244 ++++++++++++++++++-------------------
 1 file changed, 5122 insertions(+), 5122 deletions(-)
make: *** [Makefile:31: check] Error 1
```

Failed successful:

https://github.com/rzr/xcp/actions/runs/24389156947/job/71230340848# 

```
5 info: check: will pass if sources are linted, if not please lint and commit 
git diff --stat --exit-code
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,24 @@
+# YAML -*- mode: yaml; tab-width: 2; indent-tabs-mode: nil; coding: utf-8 -*-
+---
+name: check
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+  push:
+
+jobs:
+  check:
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
+
+      - name: checkout
+        # yamllint disable-line rule:line-length
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+      - name: check
+        run: make check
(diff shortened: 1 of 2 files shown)


glehmann:
I was thinking of moving this file to a more human-friendly format, like TOML.
Would that check still make sense for other files if we do that?

scripts/rpm_owners/packages.json
```diff
@@ -1,5124 +1,5124 @@
 {
-    "GConf2": {
```
ydirson, line 2: Can't we just tell `jq` that we use 4-space indent?

BTW, this operation is a reformatting rather than a linting

rzr: done, It can --indent n: 

BTW What are benefits of using 4 spaces ? Even my stoneage editor uses 2 for json by default :)

stormi: > BTW What are benefits of using 4 spaces ?

Your question implies that 4 space would be a weird choice. I don't know if there's a right answer to the way it's formulated.

rzr: well i assume that 2 spaces in the widely adopted standard for JavaScript, that's probably why jq use 2 by default, but no problem to diverge from this (maybe I'll add an .editorconfig file later)

glehmann (approved):
LGTM, we can rethink that if we ever switch to another format!

stormi (approved):
LGTM. Will merge without waiting for Yann's re-review, given the delays and the kind of PR this is.

PR #805, opened 2026-04-20
Package koji_build and import_srpm

psafont:
This should make it easier to install all tools as we pacakge them, but more importantly, these will be used by the xapi CI to make builds from PR, and test them


scripts/koji/README.md
```diff
@@ -1,11 +1,17 @@
...
+## Install koji_build
 
-~~~sh
-pip install --user -r requirements.txt
-~~~
+Run the command:
 
-to install the required dependencies.
+```sh
+uv tool install ..
```
rzr, line 10:
```suggestion
uv tool install ../..
```

Or maybe you wanted the README to be relocated in parent folder ?

uv.lock
```diff
@@ -0,0 +1,34 @@
+version = 1
```
rzr, line 1: I would commit lock file on release (tags), but since we are not tagging maybe it's better this way.

BTW what about tagging when CI is passing (end to end) our monthly update batch ?

PR #810, opened 2026-05-20
scripts: Fix `is-a-snapshot` field with underscores

AnthoineB:
With this typo, the `elif` branch will always be false and not executed, it's dead code. As it is a less common issue than the first branch of the `if`, the script will left behind this particular case but will still fix the majority of `snapshot_of` problems.

The script can still be used safely.


psafont (approved):

stormi:
We need to add a CODEOWNERS file here. No one saw this PR in @xcp-ng/os-platform-release and I only see it now because I'm starting to reach the bottom of my pile of e-mail notifications.

glehmann:
see https://github.com/xcp-ng/xcp/pull/815

PR #812, opened 2026-05-28
Scripts: enforce XCP-ng 8.3 or higher

Danp2:
This script doesn't work correctly on XCP-ng 8.2.1, so adding check for 8.3

Signed-off-by: Dan Pollak <danpollak2@gmail.com>

### DIFF
--- a/scripts/snapshot-fixer.py
+++ b/scripts/snapshot-fixer.py
@@ -27,6 +27,9 @@ xapi_db        = '/var/lib/xcp/state.db'
 xapi_db_backup = '/var/lib/xcp/state.db.snapshot_of.backup'
 xapi_db_fixed  = '/var/lib/xcp/state.db.snapshot_of.fixed'
 
+MIN_MAJOR = 8
+MIN_MINOR = 3
+
 def service_status(name):
     cmd = ['systemctl', 'is-active', name]
     r = subprocess.run(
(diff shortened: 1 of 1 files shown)


psafont (approved):

scripts/snapshot-fixer.py
```diff
@@ -181,6 +184,35 @@ def rewrite(args):
...
+    with open("/etc/xensource-inventory") as f:
+        for line in f:
+            if "=" in line:
+                k, v = line.strip().split("=", 1)
+                inventory[k] = v.strip("'")
+
+    version = inventory.get("PRODUCT_VERSION")
+
+    if not version:
+        raise RuntimeError("PRODUCT_VERSION not found")
```
olivierh-pro, line 199: Accessing a non-existing key in dict already raises an Error:

```python
>>> d = {}
>>> d["bar"]
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    d["bar"]
    ~^^^^^^^
KeyError: 'bar'
```

Maybe, `RuntimeError` is more meaningful for you, isn't it?

stormi: I'm not sure I understand what you suggest. If you suggest raising KeyError, then I think RuntimeError is equally if not more appropriate. As a user, I don't like it when python scripts raise instead of displaying a user-friendly error message, but if they must raise, let them not raise KeyError, which is the kind of exception that is deeply related to implementation details.

olivierh-pro: That's why I've asked for "meaningfulness". Your answer is crystal clear, thank you!

olivierh-pro (approved):

PR #813, opened 2026-06-05
replace poke_service on stop_xapi() systemctl stop and add 60 seconds polling loop…

Amine-hml:
replaced the poke_service() with a direct systemctl stop command and added a custom 60 second polling loop that checks the service state every 2 seconds, continuing with a warning instead of timing out fatally.

### DIFF
--- a/scripts/snapshot-fixer.py
+++ b/scripts/snapshot-fixer.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 from argparse import ArgumentParser
 import logging
 import shutil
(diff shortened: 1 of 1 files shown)


psafont (changes requested):
Not stopping xapi and still going ahead and changing the database can be catastrophic, please remove that.

Then the loop is reimplementing what `poke_service` does, the new functionality should be delivered within `poke_service`.

Please add a new optional parameter to poke_service to be able to define the timeout, and add it to the stop_xapi function with `timeout=60`.

There is functionality being removed as well, please add it back again.

Amine-hml:
Alright, thanks for the review @psafont, i'm closing this one. I'll reopen another one once i've made those modifications.

PR #820, opened 2026-06-21
Enhancements to snapshot-fixer.py

Danp2:
- Increase XAPI stop timeout to 60 secs

    Script will often timeout because XAPI doesn't stop within 15 seconds

    - Adds optional timeout parameter to poke_service function
    - Change stop_xapi function to wait 60 seconds

- Add optional --database argument

    Allows the user to specify an alternative file to be processed, which allows for examination of XAPI databases provided by the customer or captured via xen-bugtools.

- Ensure the existence of the xensource inventory file

### DIFF
--- a/scripts/snapshot-fixer.py
+++ b/scripts/snapshot-fixer.py
@@ -27,6 +27,7 @@ xapi_db        = '/var/lib/xcp/state.db'
 xapi_db_backup = '/var/lib/xcp/state.db.snapshot_of.backup'
 xapi_db_fixed  = '/var/lib/xcp/state.db.snapshot_of.fixed'
 
+bypass_checks = False
 MIN_MAJOR = 8
 MIN_MINOR = 3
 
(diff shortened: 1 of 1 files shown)


last-genius (approved):

last-genius (approved):

scripts/snapshot-fixer.py
```diff
@@ -27,6 +27,7 @@
 xapi_db_backup = '/var/lib/xcp/state.db.snapshot_of.backup'
```
glehmann: _This comment was made on the commit metadata for commit c8804a4, on the Subject line._



This commit is lacking a title. Could you add one?

glehmann: This commit seems empty, I think you can just remove it

scripts/snapshot-fixer.py
```diff
@@ -214,7 +230,10 @@ def ensure_supported_version():
         sys.exit(1)
 
 def main():
+    global xapi_db, xapi_db_backup, xapi_db_fixed, bypass_checks
```
glehmann: We usually avoid using global variables and prefer passing the values to the functions that need them

Danp2: See 3d038b8db33d71233ef924697ab8ab9ca2b0cf6f

scripts/snapshot-fixer.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
```
glehmann, line 1: _This comment was made on the commit metadata for commit c8804a4, on the Subject line._



This commit is lacking a title. Could you add one?

glehmann, line 1: This commit seems empty, I think you can just remove it

scripts/snapshot-fixer.py
```diff
@@ -202,19 +219,20 @@ def get_xcpng_version():
...
+    if (major, minor) < (cfg.min_major, cfg.min_minor):
         print(
             f"ERROR: snapshot-fixer.py requires XCP-ng "
-            f"{MIN_MAJOR}.{MIN_MINOR} or newer. "
+            f"{cfg.min_major}.{cfg.min_minor} or newer. "
             f"Detected version: {major}.{minor}"
         )
         sys.exit(1)
 
 def main():
```
glehmann, line 232: We usually avoid using global variables and prefer passing the values to the functions that need them

glehmann (approved):

glehmann (approved):

PR #823, opened 2026-07-10
scripts: Add helper script to list possibly leftover VBDs due to VM.revert regression

last-genius:
A regression was introduced to VM.revert in xapi-26.1.4-3.3, which left VBDs attached to the VM when they are not present in the reverted-to snapshot (instead of destroying such VBDs, as was done before).

VMs that were reverted to their snapshots after xapi-26.1.4-3.3 may have such "leftover" VBDs. This script can be used to detect possible cases in pools. Sadly, there is no way to automatically distinguish a VBD that was created after a VM.revert from a VBD that survived the VM.revert, so this script relies on the user to determine if the suspicious VBDs are indeed affected or not.

Therefore, tbis is not a destructive script, it just uses the API to find the affected VBDs and presents them to the user, as in this example output:

    VM e2678368-596d-1b6e-7211-9ee45f60dec1 ("Alpine minimal for CI") has
    VBDs that are not present in the snapshot it was reverted to
    (303ead66-9515-b1c8-0dc5-d33b7644b161 "Alpine minimal for CI_snapshot1"):
    xvdb (ca8610d2-76e3-6c54-4378-ee599619b4bb))

The script follows this algorithm:

1. Find all VMs that are children of snapshots (i.e. were reverted to a particular snapshot at some point in time)
2. Find VBDs that are currently present in such VMs but are not present in the snapshot

### DIFF
--- a/scripts/xapi_leaked_vbds.py
+++ b/scripts/xapi_leaked_vbds.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+
+import XenAPI
+
+def main():
+    session = XenAPI.xapi_local()
+    xapi = session.xenapi
+    try:
+        xapi.login_with_password("root", "", "0.1", "leaked_vbds_script")
+        # Get a list of snapshots
+        vms = xapi.VM.get_all_records()
+        snap_vms = {ref: vm for ref, vm in vms.items() if vm['is_a_snapshot']}
+        nonsnap_vms = {ref: vm for ref, vm in vms.items() if not vm['is_a_snapshot']}
+
+        # Dict of {vm: last_snapshot}
+        vm_snaps = {vm_ref: vm['parent'] for (vm_ref, vm) in nonsnap_vms.items() if
+                    vm['parent'] != 'OpaqueRef:NULL' and
+                    not vm['parent'].startswith('Ref:')}
+
+        # Find the snapshots the VM could have been reverted to
+        vm_reverted_snaps = {}
+        for vm_ref, snap in vm_snaps.items():
+            snaps = [snap]
+            while True:
+                new_snap = xapi.VM.get_parent(snap)
+
+                if new_snap.startswith('Ref:'):
+                    new_snap = next((vm_ref for (vm_ref, vm) in snap_vms.items()
+                                     if vm['children'] and vm['children'][0] == snap),
+                                    'OpaqueRef:NULL')
+                if new_snap == 'OpaqueRef:NULL':
+                    break
+                snaps.append(new_snap)
+                snap = new_snap
+
+            vm_uuid = xapi.VM.get_uuid(vm_ref)
+
+            # VM keeps the snapshot's install time
+            install_time = xapi.VM_metrics.get_install_time(xapi.VM.get_metrics(vm_ref))
+            snaps = [snap for snap in snaps if
+                     install_time == xapi.VM_metrics.get_install_time(xapi.VM.get_metrics(snap))]
+
+            if snaps:
+                vm_reverted_snaps[vm_ref] = snaps
+
+        for vm, snapshots in vm_reverted_snaps.items():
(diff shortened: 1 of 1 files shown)


olivierh-pro (commented):
Maybe you should wrap your code with the following pattern:

```python
def main():
    # your code here


if __name__ == "__main__":
    main()
```

What do you think?

scripts/xapi_leaked_vbds.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+import XenAPI
```
olivierh-pro, line 3: You're using an external dependency here but I can't see any requirements or something related to dependencies management (like pip or uv, or whatever).

Is it normal?

last-genius: This script is intended to be run on xcp-ng hosts, where this dependency is always available

last-genius:
> ```python
> if __name__ == "__main__":
> ```

I can, but I don't really see the point - the use-cases this guard usually protects from don't really apply here

vxgmichel (approved):
I tested this script after following the instructions from https://github.com/xcp-ng-rpms/xapi/pull/138:

> Step-by-step:
> 
>     1. VM only has xvda VBD
>     2. It's snapshotted
>     3. Another VBD is added as xvdb, pointing to VDI X
>     4. VM is reverted
>     5. VM has xvda VBD (pointing to the snapshotted VDI) and still has xvdb (pointing to VDI X)

And the extra VBD was properly detected :+1: 
```
VM 64699fa5-b5a9-b6c4-cdbd-bbbdadaf7e83 ("Alpine minimal uefi for CI") has VBDs that are not present in the snapshot it was reverted to (829b3d60-13de-1fdd-34cc-7dcda2fdd111 "Alpine minimal uefi for CI_2026-07-10T16:11:13.051Z"): xvdb (39be895b-047a-a9e4-3c72-94cad62929d4)
```

vxgmichel:
@last-genius 
> > ```
> > if __name__ == "__main__":
> > ```
> I can, but I don't really see the point - the use-cases this guard usually protects from don't really apply here

Some tools might try to import modules as part of their discovery process, and it's hard to know exactly which ones are going to run in this repository (considering that each developer might use different tools). Hence I'm +1 on @olivierh-pro suggestion (although I [already approved](https://github.com/xcp-ng/xcp/pull/823#pullrequestreview-4684168325) the PR :)

last-genius:
Moved to the `__name__ == "__main__"` pattern

olivierh-pro (approved):

rzr:
Do we want to document how to use this script in the official doc or blog post ?

last-genius:
> Do we want to document how to use this script in the official doc or blog post ?

I don't believe we've documented `snapshot-fixer.py` like this. This script is much less critical and the regression affects fewer (if any) users. I could document it somewhere in the docs though, what would be the right place?

rzr:
>  I could document it somewhere in the docs though, what would be the right place?

I think, this is the best place for this, and if not ready it can be also reused in the blog post (soon).

last-genius:
> I think, this is the best place for this, and if not ready it can be also reused in the blog post (soon).

I'm confused, what do you mean by "this"?

rzr:
> > I think, this is the best place for this, and if not ready it can be also reused in the blog post (soon).
> 
> I'm confused, what do you mean by "this"?

you said " somewhere in the docs", what about adding an article to:

https://github.com/xcp-ng/xcp-ng-org/tree/master/docs/troubleshooting

Or more precisely in:

https://github.com/xcp-ng/xcp-ng-org/tree/master/docs/troubleshooting/storage

We can discuss with reviewers, but i can help testing it.

last-genius:
Opened the PR for the docs: https://github.com/xcp-ng/xcp-ng-org/pull/499

PR #827, opened 2026-07-17
packages.json: Relocate libidn2

rzr:
Sort alphabetically


glehmann (approved):

ydirson (changes requested):
Content is OK, but a commit title should never be "fixup!" unless the intent is to squash it into another

gduperrey (approved):

PR #835, opened 2026-08-06
Add koji-helper.py: a Python script to list source packages from a Koji tag and updates

rzr:
This script is used to generated versions updates:

    uv run scripts/koji/koji-helper.py \
       list-update v8.3-ci \
       v8.3-testing v8.3-candidates v8.3-updates v8.3-base \
        | sed -e 's|\([^:]*\)\(.*\)|* `\1`\2|g' -e 's|UNKNOWN -> ||g'

It produces this kind of output (sed format to markdown):

   * `amd-microcode`: 20251203-1.1.xcpng8.3 -> 20260519-1.1.xcpng8.3
   * `perl-Archive-Tar`: 1.92-3.el7

### DIFF
--- a/scripts/koji/koji-helper.py
+++ b/scripts/koji/koji-helper.py
@@ -52,6 +52,8 @@ def list_source_packages(session, tag, latest=False):
 
     Returns a dict mapping package name to its info:
         - nvr: name-version-release string
+        - version: package version
+        - release: package release
     """
     # Get the tag info
     tag_info = session.getTag(tag, strict=True)
@@ -64,7 +66,9 @@ def list_source_packages(session, tag, latest=False):
         name = build["package_name"]
         if name not in results:
             results[name] = {
-                "nvr": build["nvr"]
+                "nvr": build["nvr"],
+                "version": build["version"],
+                "release": build["release"],
             }
 
     return tag_info, results
@@ -79,7 +83,11 @@ def list_update_source_packages(session, to_tag, from_tag=None, latest=False):
 
     Returns a dict mapping package name to its info:
         - nvr: name-version-release string in to_tag
+        - version: package version in to_tag
+        - release: package release in to_tag
         - base_nvr: name-version-release string in from_tag (if provided)
+        - base_version: package version in from_tag
+        - base_release: package release in from_tag
     """
     to_tag_info, to_packages = list_source_packages(session, to_tag, latest=latest)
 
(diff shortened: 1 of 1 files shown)


scripts/koji/koji-helper.py
```diff
@@ -0,0 +1,247 @@
...
+    nvr: str
+    version: str
+    release: str
+    base_version: NotRequired["PackageInfo | None"]
+
+
+Packages = dict[str, PackageInfo]
+
+
+def get_session(args: argparse.Namespace) -> koji.ClientSession:
```
glehmann, line 46: The script can't connect on my computer.

In `pkg_in_pipe.py`, we do:

```python
            # open koji session
            config = koji.read_config("koji")
            KOJI = koji.ClientSession('https://kojihub.xcp-ng.org', config)
            KOJI.ssl_login(config['cert'], None, config['serverca'])
```

scripts/koji/koji-helper.py
```diff
@@ -0,0 +1,247 @@
+#!/usr/bin/env python3
```
glehmann, line 1: The script should be executable.

scripts/koji/koji-helper.py
```diff
@@ -0,0 +1,247 @@
...
+
+    args = parser.parse_args()
+
+    if not args.command:
+        parser.print_help()
+        sys.exit(1)
+
+    session = get_session(args)
+
+    try:
```
glehmann: You can just remove this try/catch, and let the exception propagate

scripts/koji/koji-helper.py
rzr: This change seems to work too (at least with koji 1.36.1)

glehmann (approved):

scripts/koji/koji-helper.py
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Philippe Coval <philippe.coval@vates.tech>
+#
+# SPDX-License-Identifier: MIT
```
vxgmichel: Is this still necessary? The repo already has a license.

vxgmichel (approved):
In the PR description you mentioned this sed:
```
[...] | sed -e 's|\([^:]*\)\(.*\)|* `\1`\2|g' -e 's|UNKNOWN -> ||g'
```

Shouldn't this formatting be part of script instead?

LGTM otherwise 👍

scripts/koji/koji-helper.py
rzr: ok then let me remove it, until we decide to track all files (in other PR)

rzr:
Ok let's merge, the reformatting can be done in later PR if there is a need for.

PR #847, opened 2026-09-18
 [pkg_in_pipe] Try all the koji tagged builds in order until we find a source

vxgmichel:
It might happen that the previous tag build for a package does not have its source field filled in with the corresponding git commit (this typically happens for security updates that did not go through the standard release procedure). By trying again with the build before it until we find a source, we limit the amount of PRs and Plane cards that will be included in the reports (if no source is provided by `find_previous_build_commit`, all the PRs and cards are listed).

In order to avoid reporting extra PRs and plane cards, those missing sources can also now be listed in the `missing_sources.json` dictionary.


glehmann (commented):
I'd rather avoid the json format, which we too often get wrong when adding an element, because we can't have a comma `,` on the last element of the list.
Could you change it for toml?

vxgmichel:
> Could you change it for toml?

Sure, done here: 22e1dec1746bc555e516c2c259dbc071ed1b9cd6

glehmann (approved):

rzr (approved):

